Issue #100, opened 2018-09-17
Map symbology does not update when the last remaining inspection is deleted


mikewilburn:
The tree feature's details which depict the `Condition` of the tree also do not update.
This was one piece of feedback that came out of Sarat's testing of the app (result from test case 4.5).

Steps to repro:
1. Find a tree which only has one inspection record left.
2. Take note of the tree's condition and map symbol color, then delete the inspection.
3. Notice that neither the map symbol updates nor does the `Condition` attribute of the feature get cleared.

esreli:
@mikewilburn A PR has opened that solves this.

esreli:
@mikewilburn This change has been merged into master. Please confirm it has been fixed.

mikewilburn:
Verified on master.

PR #122, opened 2018-09-29
Improves app custom behavior: Tree symbology updates despite zero inspections.

esreli:
Changes update symbology custom behavior to update symbology when there are zero inspections for a tree.

To confirm this PR addresses the issue:

1. Launch app. (Using Trees of Portland, default)
2. Identify a tree on the map, ensure that tree has 1+ inspections.
_Note the color of the tree symbology_
3. Delete all of the inspections for that tree.
4. Return to the map view.

At this point the tree symbology should be black, representing no inspection.

If you add an inspection(s) to the tree, the symbology should reflect the color of the most recent inspection.

### DIFF
--- a/data-collection/data-collection/Custom Behavior/CustomBehavior_TreeSymbology.swift
+++ b/data-collection/data-collection/Custom Behavior/CustomBehavior_TreeSymbology.swift
@@ -56,12 +56,10 @@ func updateSymbology(withTreeManager treeManager: PopupRelatedRecordsManager, co
     if let newestInspection = inspectionsManager.relatedPopups.first,
         let newestInspectionFeature = newestInspection.geoElement as? AGSArcGISFeature,
         let newestInspectionFeatureTable = newestInspectionFeature.featureTable as? AGSArcGISFeatureTable,
-        newestInspectionFeatureTable.tableName == "Inspections",
-        newestInspectionFeature.attributes[conditionKey] != nil,
-        newestInspectionFeature.attributes[dbhKey] != nil {
+        newestInspectionFeatureTable.tableName == "Inspections" {
          // Update the condtion and dbh of the tree reflecting those of the newest inspection.
-        treeFeature.attributes[conditionKey] = newestInspectionFeature.attributes[conditionKey]
-        treeFeature.attributes[dbhKey] = newestInspectionFeature.attributes[dbhKey]
+        treeFeature.attributes[conditionKey] = newestInspectionFeature.attributes[conditionKey] ?? NSNull()
+        treeFeature.attributes[dbhKey] = newestInspectionFeature.attributes[dbhKey] ?? NSNull()
     }
     // Unless there is no newest inspection.
     else {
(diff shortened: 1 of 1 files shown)


esreli:
References this issue:
https://github.com/Esri/data-collection-ios/issues/100

data-collection/data-collection/Custom Behavior/CustomBehavior_TreeSymbology.swift
```diff
@@ -35,49 +35,41 @@ func updateSymbology(withTreeManager treeManager: PopupRelatedRecordsManager, co
...
+    if let newestInspection = inspectionsManager.relatedPopups.first,
         let newestInspectionFeature = newestInspection.geoElement as? AGSArcGISFeature,
         let newestInspectionFeatureTable = newestInspectionFeature.featureTable as? AGSArcGISFeatureTable,
-        newestInspectionFeatureTable.tableName == "Inspections"
-        else {
-            print("[Error: Update Symbology] could not find inspections table.")
-            completion()
-            return
+        newestInspectionFeatureTable.tableName == "Inspections",
+        newestInspectionFeature.attributes[conditionKey] != nil,
```
philium: It would be best to do the lookup once:
```swift
if …,
    let condition = newestInspectionFeature.attributes[conditionKey],
    let dbh = newestInspectionFeature.attributes[conditionKey] {

    treeFeature.attributes[conditionKey] = condition
    treeFeature.attributes[dbhKey] = dbh
} else { … }
```

data-collection/data-collection/Custom Behavior/CustomBehavior_TreeSymbology.swift
```diff
@@ -35,49 +35,41 @@ func updateSymbology(withTreeManager treeManager: PopupRelatedRecordsManager, co
...
-    }
-    
-    guard newestInspectionFeature.attributes[conditionKey] != nil, newestInspectionFeature.attributes[dbhKey] != nil else {
-        print("[Error: Update Symbology] inspection attributes doesn't contain condition or dbh keys.")
-        completion()
-        return
+    // Unless there is no newest inspection.
+    else {
+         // Update the condtion and dbh of the tree to reflect a missing inspection.
+        treeFeature.attributes[conditionKey] = NSNull()
```
philium: Is a key's value being `NSNull()` semantically different from its value being `nil`? I don't see `NSNull` used anywhere else. Is this an `ArcGIS`-specific thing?

esreli: It's partially an `ArcGIS`-specific thing. These attributes can be nil but setting the attribute value `nil` removes the dictionary entry. When this attributes dictionary is persisted, changes to removed attributes aren't performed and is thus ignored. Conversely, setting the `NSNull()` nil pointer explicitly tells ArcGIS to set the value for key to nil and keeps the entry in place.

To see what i'm talking about, feel free to swap `NSNull()` for `nil` and follow these steps:

1. Launch App using Trees of Portland
2. Identify a tree on the map.
3. Remove all the inspections related to that tree

Running the debugger, you can see that not just the value has been set nil but the whole entry removed from the dictionary. You will notice after the changes are persisted that the symbology has not updated to reflect black and that if you interrogate the feature's attributes dictionary the old value persists.

PR #124, opened 2018-10-05
Location Display Improvements

esreli:
Improves how location display is started and how the app handle location auth status changes.

To verify this change, follow these **2** scenarios:

**A**
1. Delete Data Collection app from device
2. Launch Data Collection app
3. Grant Data Collection app access to device location
_Location Display should be visible_
4. Switch to Settings app, remove authorization to device location
5. Switch back to Data Collection app
_Location Display should not be visible_
6. Switch to Settings app, grant authorization to device location
7. Switch back to Data Collection app
_Location Display should be visible_

**B**
1. Delete Data Collection app from device
2. Launch Data Collection app
3. Do not grant Data Collection app access to device location
_Location Display should not be visible_
4. Switch to Settings app, grant authorization to device location
5. Switch back to Data Collection app
_Location Display should be visible_


philium:
FYI: There is an alternative to deleting the app. If you'd prefer to keep the app's data, you can use Settings > General > Reset > Reset Location & Privacy.

PR #127, opened 2018-10-10
Modifies origin directly rather than through a helper method

philium:
Also removes related unused helper methods on `CGRect` and `CGPoint`.


philium:
> Is there a reason you chose to leave in func modified(x xDelta: CGFloat) -> CGPoint ?

I checked that `CGPoint.modified(y:)` was being used and just assumed the other one was being used as well.

esreli:
I left it because it felt more balanced, despite not being used. If the app provides an API for interacting with CGPoint in such a way for `y`, it should do the same for `x`. I would still suggest it stays. Was just curious why you hadn't nixed it like you did the other APIs.

philium:
Agreed.

PR #129, opened 2018-10-12
improves activity bar setting colors and animations

esreli:
This improves the activity bar API as well removing the redundant update for new color call.

### DIFF
--- a/data-collection/data-collection/Custom Views/ActivityBarView.swift
+++ b/data-collection/data-collection/Custom Views/ActivityBarView.swift
@@ -19,8 +19,7 @@ class ActivityBarView: UIView {
     
     init(mapView: AGSMapView, colors:(a: UIColor, b: UIColor)) {
         self.mapView = mapView
-        self.colorA = colors.a
-        self.colorB = colors.b
+        self.colors = colors
         
         super.init(frame: .zero)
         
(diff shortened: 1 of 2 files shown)


data-collection/data-collection/Custom Views/ActivityBarView.swift
```diff
@@ -66,28 +59,28 @@ class ActivityBarView: UIView {
...
     private func startProgressAnimation() {
         isAnimating = true
         alpha = 1.0
         isHidden = false
-        UIView.animate(withDuration: 0.1, animations: { [weak self] in
-            self?.alpha = 1.0
-        }, completion: { [weak self] (completion) in
+        backgroundColor = colors.a
+        
+        let swapAnimation: (Bool) -> Void = { [weak self] (_) in
```
philium: I would recommend checking the value of the given Boolean parameter and bail if the animation hasn't finished.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
```diff
@@ -216,8 +216,9 @@ class MapViewController: UIViewController {
         }
 
         let workModeChange: AppContextChange = .workMode { [weak self] workMode in
-            self?.activityBarView.colorA = (workMode == .online) ? UIColor.primary.lighter : UIColor.offline.lighter
-            self?.activityBarView.colorB = (workMode == .online) ? UIColor.primary.darker : UIColor.offline.darker
+            let a = (workMode == .online) ? UIColor.primary.lighter : UIColor.offline.lighter
```
philium: It seems like this could be simplified quite a bit:
```swift
guard let self = self else { return }
let color: UIColor = workMode == .online ? .primary : .offline
self.activityBarView.colors = (a: color.lighter, b: color.darker)
```
Let me know what you think.

esreli: It's great.

esreli:
Thank you for your review @philium, would you have another look?

Issue #139, opened 2018-10-24
Collect new inspection button still appears even if feature service doesn't participate in 1:M relationship


mikewilburn:
This came up while testing with a custom web map which contains only a single feature layer - `59ad551de0f042f088ee99ea26321b96`.

Running the app against that web map, you'll notice that identifying a tree point still includes the `+` button to collect a new 1:Many related record, despite the fact that the layer does not participate in a relationship. We should conditionally remove the button if no such relationship exists.

The good news is that we have pretty good error messaging for this case:
![image](https://user-images.githubusercontent.com/25207711/47398163-b6f37c80-d6e7-11e8-9035-d478581626f0.png)

esreli:
@mikewilburn changes have been merged to master. Can you confirm the app works as expected?

mikewilburn:
Verified on master that the `+` button no longer shows up if the identified feature does not participate in a 1-to-Many relationship.

![image](https://user-images.githubusercontent.com/25207711/47581356-01eadb00-d906-11e8-9f5b-700c7fcac7e9.png)

PR #142, opened 2018-10-26
Add Related Record Button Bugfix

esreli:
Defaults add related record button to hidden- unless first one to many related record table can add feature.

See issue: https://github.com/Esri/data-collection-ios/issues/139

### DIFF
--- a/data-collection/data-collection/View Controllers/Map View Controller/MapViewController+SmallPopupView.swift
+++ b/data-collection/data-collection/View Controllers/Map View Controller/MapViewController+SmallPopupView.swift
@@ -61,10 +61,11 @@ extension MapViewController {
                 self?.relatedRecordsNLabel.text = fallbackPopupManager?.nextDisplayFieldStringValue(fieldIndex: &fallbackIndex)
             }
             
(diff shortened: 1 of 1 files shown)


data-collection/data-collection/View Controllers/Map View Controller/MapViewController+SmallPopupView.swift
```diff
@@ -61,7 +61,7 @@ extension MapViewController {
                 self?.relatedRecordsNLabel.text = fallbackPopupManager?.nextDisplayFieldStringValue(fieldIndex: &fallbackIndex)
             }
             
-            self?.addPopupRelatedRecordButton.isHidden = false
+            self?.addPopupRelatedRecordButton.isHidden = true
```
philium: What would you think of moving this into an `else` clause of the following `if` statement? That was the `isHidden` property would only be set once per execution of this closure. The way it is now, the button *could* be hidden and then unhidden.

esreli: good idea

Issue #141, opened 2018-10-25
Providing app credentials to access a private resource retains those credentials even after logging out


mikewilburn:
Steps to repro:
- Launch the app with web map item id `1300bf23f97f4a2bb6ede795dfab8b72` (this is a private web map that contains private layers)
> Notice that when the map loads, you're prompted to log in
- Log in with your credentials
> Notice that the map loads, complete with a feature layer this time
> You're able to interact with the map, identify features, view their attributes, and even edit them
- Go to the hamburger menu and log out
> The map remains loaded and all features still exist in the map view & are identifiable

**Expected behavior:** logging out should remove access to the private resource

esreli:
@mikewilburn this change has been merged into `master`. Can you confirm it's been fixed?

mikewilburn:
Verified that this is working as it should.
After logging out of the app, the authentication page re-appears prompting the user to provide their credentials.
Clicking `CANCEL` or `Done` on this page returns to the app, but the next action to interact with the map (pan, zoom, or click to identify a feature) will re-prompt the authentication page.
Marking this closed.

esreli:
I've determined this issue was due to a bug. Will be opening a PR to fix this bug.

mikewilburn:
In some ways, this works even better than before. For instance, the act of logging out immediately prompts the Log in dialog. Clicking `Cancel` returns to the map view but with the message "Could not load map. <Reload Map>" clearly displayed. Clicking `Reload Map` re-prompts the authentication challenge or Log in dialog to appear again.
Verified that this bug fix is working as expected.

PR #143, opened 2018-10-26
Removes credentials for every AGSServiceFeatureTable upon logout

esreli:
This PR references this issue:
https://github.com/Esri/data-collection-ios/issues/141

The steps to repro are contained within the issue.


data-collection/data-collection/Extensions/ArcGIS/AGSMap/AGSMap+GeodatabaseFeatureTable.swift
```diff
@@ -26,4 +26,10 @@ extension AGSMap {
         return tables.compactMap { return ($0 as? AGSGeodatabaseFeatureTable) } +
             operationalLayers.compactMap { return (($0 as? AGSFeatureLayer)?.featureTable as? AGSGeodatabaseFeatureTable) }
     }
+    
+    var allServiceTables: [AGSServiceFeatureTable] {
+        
+        return tables.compactMap { return ($0 as? AGSServiceFeatureTable) } +
```
nixta: I think we should also include the basemap here, considering the `baseLayers` and `referenceLayers` . The `AGSRemoteResource` protocol might be useful here.

PR #146, opened 2018-10-26
Refreshes MapViewController for newly set portal

esreli:
When a new portal is set, the map view controller refreshes it's `mapViewMode` and `currentPopup`.

### DIFF
--- a/data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
+++ b/data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
@@ -226,6 +226,8 @@ class MapViewController: UIViewController {
         }
         
         let currentPortalChange: AppContextChange = .currentPortal { [weak self] portal in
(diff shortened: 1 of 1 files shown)


data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
```diff
@@ -226,6 +226,8 @@ class MapViewController: UIViewController {
         }
         
         let currentPortalChange: AppContextChange = .currentPortal { [weak self] portal in
+            self?.mapViewMode = .defaultView
```
philium: Since you're now unwrapping `weak self` multiple times, I would strongly suggestion doing it once at the beginning of the closure. And Swift 4.2 made it even easier:
```swift
guard let self = self else { return }
```

esreli: I agree, though i'm saving that for a wide-spread PR.

PR #147, opened 2018-10-30
Bug fix for maps requiring credential

esreli:
I've determined the behavior the app was experiencing was due to an app bug (as reported in this issue: https://github.com/Esri/data-collection-ios/issues/141)

When a user logged out, the `AppContext` built a new `AGSPortal` and published the new portal to the rest of the app. The problem was, just because a new `AGSPortal` was built, does not mean that a corresponding new `AGSMap(item: AGSPortalItem)` was built as well. If the app was in online mode, a reference persisted to the map built using the previous portal, as well as the **credentials**. Those credentials would then get re-entered into the credential cache. 

Now, when a new portal is loaded- if successful *and* app work mode is online, a new `AGSMap` is built using a `AGSPortalItem` derived from the new `AGSPortal`. If unsuccessful *and* app work mode is online, the current map is set to nil.


esreli:
@nixta i'll confirm.

Issue #120, opened 2018-09-28
Consider: Improve accessibility of app.


esreli:
Start with visual elements & move from there.

https://developer.apple.com/accessibility/ios/

esreli:
v1.1 introduces basic accessibility of visual elements.

PR #148, opened 2018-10-31
Improves Accessibility

esreli:
This PR enhances the accessibility of the app through dynamic font, content size and higher contrast colors.

When solving for dynamic font size, a number of concerns are raised. At a font that is largest, can all of the content be read? Will certain UI elements cover others?

With colors, other concerns are raised. How does the app look for those with varying degrees of color blindness?

**Note** As a result of the need to change the app's colors, I decided to lump the [accessibility issue](https://github.com/Esri/data-collection-ios/issues/120) with the [global appearance issue](https://github.com/Esri/data-collection-ios/issues/133). This way, I could change colors throughout the app as I worked on color accessibility.

To see the app at various levels of dynamic font size, open the **Accessibility Inspector** and attach it to the process running the app. Then, click the cog icon and adjust the **Font size** slider. _Note: UIButtons don't automatically adjust for content size changes like labels do_. The best way to test UIButtons at various sizes is to re-run the application with a different font size.

## Testing considerations

**AppContainerViewController**
- Navigation Bar Items (press and hold)

**MapViewController**
- Select Point View
- Select Offline Extent View
- Sliding Label (Could not identify results)
- Small pop-up view
- Reload map map view mode

**DrawerViewController**
- Drawer buttons
- Drawer width adjusts once "accessibility threshold" reached

**PopupRelatedRecordsViewController**
- Table cells
- Table headers
- Navigation Bar Items (press and hold)

**JobStatusViewController**
- Button
- Status Label


data-collection/data-collection/Custom Views/SlideNotificationView.swift
```diff
@@ -56,21 +56,31 @@ class SlideNotificationView: UIView {
...
+        // Build a reference to the top anchor constraint
         topSlideConstraint = top
+
+        // Add a bottom anchor constraint that is equal to the superview's *top anchor* with a lower priority.
+        // This anchor will become active when `topSlideConstraint` is disabled.
+        let bottom = label.bottomAnchor.constraint(equalTo: topAnchor, constant: 0.0)
+        bottom.priority = .defaultHigh
+        
+        // Set a height constraint.
+        let height = NSLayoutConstraint(item: label, attribute: .height, relatedBy: .equal, toItem: self, attribute: .height, multiplier: 1.0, constant: 0.0)
```
philium: What does this constraint do? The label should have an intrinsic content height, so it seems odd to be constraining its height.

data-collection/data-collection/View Controllers/Drawer View Controller/DrawerViewController.swift
```diff
@@ -185,7 +186,7 @@ class DrawerViewController: UIViewController {
         
         if let currentUser = user {
             
-            let fallbackProfileImage = UIImage(named: "MissingProfile")!.withRenderingMode(.alwaysOriginal).circularThumbnail(ofSize: 36, stroke: (color: .loginLogoutNormal, weight: 1))
+            let fallbackProfileImage = UIImage(named: "MissingProfile")!.withRenderingMode(.alwaysOriginal).circularThumbnail(ofSize: 36, stroke: (color: #colorLiteral(red: 1, green: 1, blue: 1, alpha: 1), weight: 1))
```
philium: This is just `UIColor.black`, correct?

data-collection/data-collection/AppColors.swift
```diff
@@ -18,30 +18,40 @@ import UIKit.UIColor
 // These colors are used throughout the application, when colors are generated at runtime.
 extension UIColor {
     
-    static let primary: UIColor = #colorLiteral(red:0.66, green:0.81, blue:0.40, alpha:1.00)
+    // MARK: General
     
-    static let offline: UIColor = .darkGray
+    // The general of the app and color symbolizing online work mode.
+    static let primary = #colorLiteral(red: 0.3783819885, green: 0.6019210188, blue: 0.2394252939, alpha: 1)
```
philium: Did you consider putting the custom colors in the asset catalogue? That is the best practice these days. It lets you reference custom colors by name in both Interface Builder and source code. You could still have these properties on `UIColor` that just return the named color.

esreli: Good suggestion, I'll take it!

data-collection/data-collection/View Controllers/Drawer View Controller/DrawerViewController.swift
esreli: This is just `UIColor.white`.

philium: Yeah! That one. Did you use the color literal just so that the actual color is visible in the source code?

esreli: I did.

data-collection/data-collection/Custom Views/SlideNotificationView.swift
esreli: This is used to "push" the view's height based on the height of the label. Now that you raise this question, it might be operating in inverse from it's intent. But removing it causes the slide label to not appear whatsoever.

I'm open to suggestions.

philium: I just took a look at `SlideNotificationView` and I think I understand better how it works. I didn't realize that the notification view is just an invisible container for the label and that it is the label that is being animated, not the notification view itself. I would suggestion constraining the label in the notification view how you want and then animating the notification view itself. That could be done a couple different ways:

- By resizing the notification view. Have a height constraint with a constant of `0` that is enabled to hide the notification, disabled to show it. You would also want to give the label's top constraint a priority of less than `required` so that it can safely break and the label would slide up along with the bottom of the notification view.
- By moving the notification view. This should probably be done in the view controller (or whoever owns the notification view) as it is the superview's responsibility to position its subviews. You would need two mutually exclusive constraints: one that constrains the top of the notification view to the appropriate place and another that constrains the bottom of the notification view to that same place. To show the notification, you simply enable the first and disable the second. To hide it, you do the opposite.

RELEASE.md
```diff
@@ -2,7 +2,7 @@
 
 ## Major Features and Improvements
 
-- UI elements dynamic font sizing is introduced to improve accessibility.
+- UI elements dynamic type is introduced to improve accessibility.
```
philium: I think Dynamic Type is a proper noun, so I would suggest capitalizing it.

data-collection/data-collection/Custom Views/SlideNotificationView.swift
esreli: Thanks @philium. A few comments.

> I would suggestion constraining the label in the notification view how you want and then animating the notification view itself.

I wonder, at that point why would I bother with a `SlideNotificationView` at all? What you're suggesting is no different that dropping in a `UILabel` and allowing the view controller to manage the animations. 

The whole point of this view is to offload that responsibility from the view controller. That way the view controller only needs to care about specifying when to present a message and what to say.

> ... it is the superview's responsibility to position its subviews.

This is why the `SlideNotificationView` contains and manages a `UILabel` subview.

> You would need two mutually exclusive constraints: ...

This is the same technique used by the view now, is it not?

esreli: I actually think this technique works quite well and accomplishes a-lot of the goals you've suggested in this comment.

philium: Good questions! 

> why would I bother with a SlideNotificationView at all?

The notification view currently constrains the height of the label to its own height, which is slightly larger than the intrinsic height of the label. This effectively insets the label's text. It is possible to create a `UILabel` subclass with an edge insets property that insets the label's text and reports the adjusted intrinsic content size (I've done it before), but it is easier to just constrain the top and bottom of the label to a superview with a constant for insetting the label. That way the intrinsic height of the content view is determined by the intrinsic height of the label plus the top and bottom constants.

> This is the same technique used by the view now, is it not?

It is, just applied to the notification view itself rather than the label. That way the intrinsic content size of the label could be used, allowing it to be adaptable to the current content size category (if you so choose).

I assumed that the label was hiding behind the navigation bar, but it appears that the notification view is clipping its content. I suppose that works. It seems a bit odd to move the label by adjusting the constant of the top constraint. The danger with that method is that it is fragile--it breaks if the view's height changes. I don't see how that can happen for the notification view, but it is necessary to think about that. Disabling the top constraint and constraining the bottom to the top has the same affect, but is less fragile. It works regardless of whether the height of the view changes. Does that make sense? Just something to think about.

philium: To be clear, I'm just trying to brainstorm ways to take advantage of the intrinsic content size of the label rather than hardcoding the height of the notification view.

esreli: Thank you @philium, a few notes.

> To be clear, I'm just trying to brainstorm ways to take advantage of the intrinsic content size of the label rather than hardcoding the height of the notification view.

I'm curious why you feel the need to take advantage of the intrinsic content size? 

The design for this element asks the view to stretch from edge to edge of the screen and to have a height large enough to be noticed passively. Inherently, this design requires that we ignore the intrinsic content size in favor of a larger height and full width that is determined using auto layout, treating this view like we would any other view that doesn't have an intrinsic content size (like any old `UIView`). This is desired because I want the view controller to tell the slide label view how large it should be, with the intrinsic content size of the label used as a fallback in case the view needs to *expand*, a behavior that we encounter when dynamic type increases the label's font size. 

> That way the intrinsic height of the content view is determined by the intrinsic height of the label plus the top and bottom constants.

Remember, the sliding label's background color is used. If I were to follow your technique, I would need to subview the label within in another view.

>  It seems a bit odd to move the label by adjusting the constant of the top constraint.

Have another look, this is not what's happening. No constants are modified.

> Disabling the top constraint and constraining the bottom to the top has the same affect, but is less fragile. 

Have another look, what you're recommending here is already in place.

> but is less fragile. 

In fact, the technique that I have implemented here holds up under duress. I've even tried inserting a quite pushy view in testing. It accurately slides the label on and off the screen though does shrink the text a bit to make up for the quite pushy neighboring view. 

I encourage you to try it for yourself.

I see the validity of leveraging the intrinsic content size, for height, but your technique does not account for other requirements of this view.

philium: @esreli Thank you. I must have been looking at an old version. I was browsing the source through Xcode, not looking at the diff in the PR.

Issue #152, opened 2018-11-06
Set new initial viewpoint to map view when switching from working online to offline


mikewilburn:
When a user has an offline map and switches from working online to working offline, it's possible for the offline map to open at an unknown location, leaving the user swimming in a lattice of grey "no data" without any indication of scale or whereabouts.

This is because the present behavior has deterministic rules that attempt to preserve the user's online map extent as best as possible.

If this online map view extent is nowhere near the location of the already-downloaded offline map, this can make it exceptionally difficult to navigate to that area. Even worse, if the user has altogether forgotten where the offline map exists, such behavior could make it altogether impossible to locate the area covered by the offline map.

I propose:
- If the online extent intersects the offline map extent, preserve the extent at the closest scale level to the available offline map
- else set the map view to the centerpoint of the offline map at the coarsest level of detail at which it's available (smallest scale) to provide the most amount of location context

mikewilburn:
@esreli that is just one recommendation as to how we might go about solving this. There are bound to be many other valid solutions. Just didn't want to log a problem without offering a possible solution :)

mikewilburn:
@esreli There are still a few circumstances in which this isn't working.

I'm guessing a bit here, but I think this logic breaks down because the _offlineMap_'s scale can never be smaller than the coarsest LOD at which the offline map is available. So while that intersection evaluation is probably written soundly, when you:
> set _visibleArea_ to _currentExtent_

some of the extent-setting behavior might be getting overridden by the centerpoint of the _currentExtent_ being preserved, but at a sometimes greatly zoomed in scale at which the offline map is well outside of the _visibleArea_. This means you're dropped into this lattice of grey and would only be able to navigate to the offline map if you knew precisely in which direction to pan and by how much.

I can show you more of this in person, because that's pretty hard to decipher in writing.

mikewilburn:
By contrast, it looks like the .NET app permits zooming out beyond the initial extent that was used to capture the offline map:
![image](https://user-images.githubusercontent.com/25207711/48233920-0e424f80-e36c-11e8-85c4-e8927d12e604.png)

mikewilburn:
If I'm understanding correctly here, I think just **a)** would be sufficient. I think it would be fine to allow the user to zoom out as the .NET version of the app does.

esreli:
Sounds good.

esreli:
With this technique I will note that for any map that is already downloaded, the scale will remain at what is was specified at the time of the download.

[...]

PR #153, opened 2018-11-06
Ensures when offline map loads that it is at least partially visible.

esreli:
This PR addresses this enhancement request:
https://github.com/Esri/data-collection-ios/issues/152

### DIFF
--- a/data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Map.swift
+++ b/data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Map.swift
@@ -29,40 +29,49 @@ extension MapViewController {
         
         let loadCompletion: ((Error?) -> Void)? = { [weak self] (error) in
             
+            guard let self = self else { return }
+            
             guard error == nil else {
                 let error = (error! as NSError)
 
                 print("[Error: Map Load]", "code: \(error.code)", error.localizedDescription)
                 
                 if AGSServicesErrorCode(rawValue: error.code) == .tokenRequired, !appContext.isLoggedIn {
-                    self?.present(loginAlertMessage: "You must log in to access this resource.")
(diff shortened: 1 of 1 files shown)


data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Map.swift
```diff
@@ -29,40 +29,49 @@ extension MapViewController {
...
+            self.delegate?.mapViewController(self, didUpdateTitle: map.item?.title ?? "Map")
             
             guard let operationalLayers = map.operationalLayers as? [AGSFeatureLayer] else {
-                self?.delegate?.mapViewController(self!, shouldAllowNewFeature: false)
+                self.delegate?.mapViewController(self, shouldAllowNewFeature: false)
                 return
             }
             
-            AGSLoadObjects(operationalLayers, { [weak self] (_) in
+            AGSLoadObjects(operationalLayers, { (_) in
```
philium: This closure should still weakly capture `self`.

esreli:
@philium would you have another look?

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Map.swift
esreli: I wasn't under the impression this was needed. What am I missing? The retain count remains the same without capturing self weakly.

philium: The issue is that `completion` is an escaping closure. The default is to strongly capture `self`. But by doing so, the (potentially long running) asynchronous operation prevents `self` from being deallocated in the event that `self` is no longer needed. I suppose that may not make a difference if the map view controller is never deallocated during the lifetime of the app, but it is good practice to weakly capture `self` when performing asynchronous operations.

PR #160, opened 2018-11-29
Introduces RichPopup and reorganizes project accordingly.

esreli:
This is a fairly lofty PR and I hope you're not daunted by it. My suggestion would be to evaluate various components that I identify here, as a whole rather than [files changed diff lines].

Broadly speaking, this PR sets out to formally delineate pop-up **reading** from **editing**.

I've introduced `RichPopup`, which contains a single property, `relationships`, an instance of `Relationships`, a data structure that handles loading (via `AGSLoadable`) and organizing related records.

Now, an instance of `RichPopup` can load relationships:

```swift
richPopup.relationships!.load({})
``` 
and check load status of relationships:

```swift
richPopup.relationships?.loadStatus
```

A `Relationships` (note the plural form) object maintains collections of related records. A `Relationship` (note the singular form) object represents a single relationship between two feature tables and is loaded via `AGSLoadable`. A `Relationship` is not intended to be used directly but instead to be subclassed, which the app does via `OneToManyRelationship` and `ManyToOneRelationship`. More relationship types can be added down the line, easily (such as one-to-one and many-to-one). 

A `Relationship` subclass is a loadable **model** of related records for a specific relationship between and places the onus of actually relating the records into the `RichPopupManager` (née `PopupRelatedRecordsManager`).

This challenge opened up a series of app-wide changes, particularly in `MapViewController` and in `RichPopupViewController` (née `PopupRelatedRecordsViewController`) inspiring some cleanup up of this view controller as a whole.

I've also identified a number of other areas for improvement along the way, including a more clear and unified approach to `AppErrors`.


data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupViewController+EditingSession.swift
```diff
@@ -0,0 +1,350 @@
...
+
+                    guard let self = self else { return }
+                    
+                    if let error = error {
+                        self.present(simpleAlertMessage: error.localizedDescription)
+                        completion?(false)
+                        return
+                    }
+                    
+                    completion?(true)
```
philium: Consider putting this in an `else` clause and getting rid of the explicit `return` on line `111`.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupViewController+EditingSession.swift
```diff
@@ -0,0 +1,350 @@
...
+            
+            return
+        }
+        
+        do {
+            try parentPopupManager.update(popup)
+        }
+        catch {
+            completion(error)
+            return
```
philium: I was wondering if it would be possible to eliminate the explicit `return` and came up with this:

```swift
let thrownError: Error?
do {
    try parentPopupManager.update(popup)
    thrownError = nil
} catch {
    thrownError = error
}
completion(thrownError)
```

I don't know if that's any better. Feel free to take it or leave it.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupViewController+EditingSession.swift
```diff
@@ -0,0 +1,350 @@
...
+        featureTable.performEdit(feature: feature, completion: { [weak self] (error) in
+            
+            guard let self = self else { return }
+            
+            if let error = error {
+                completion(error)
+                return
+            }
+            
+            self.checkIfShouldPerformCustomBehavior() { completion(nil) }
```
philium: Consider putting this in an `else` clause.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupViewController+TableViewDelegate.swift
```diff
@@ -142,9 +146,16 @@ extension RelatedRecordsPopupsViewController {
 
         // Add delete action, if can delete.
         if let feature = childPopup.geoElement as? AGSArcGISFeature, let featureTable = feature.featureTable as? AGSArcGISFeatureTable, featureTable.canDelete(feature) {
+            
             let deleteAction = UITableViewRowAction(style: .default, title: "Delete") { [weak self] (action, indexPath) in
-                self?.present(confirmationAlertMessage: "Are you sure you want to delete this \(childPopup.title ?? "record")?", confirmationTitle: "Delete", confirmationAction: { [weak self] (_) in
```
philium: I actually think this was fine. You were only referencing `weak self` once in each scope, so there isn't an issue with `self` going away in between retrievals of it.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/RichPopup.storyboard
```diff
@@ -22,8 +22,15 @@
                             <outlet property="delegate" destination="QbY-UE-L23" id="wOR-iB-9yQ"/>
                         </connections>
                     </tableView>
-                    <navigationItem key="navigationItem" title="Popup" id="c2z-4c-LO6"/>
+                    <navigationItem key="navigationItem" title="Popup" id="c2z-4c-LO6">
+                        <barButtonItem key="rightBarButtonItem" title="Edit" id="Wnf-HI-Nv9">
```
philium: Did you consider using `UIViewController.editButtonItem`?

data-collection/data-collection/App Errors/InvalidOperationError.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+
+let InvalidOperationError = NSError(domain: "\(appBundleID).AppError.InvalidOperation", code: AppErrorBaseCode.InvalidOperation.rawValue, userInfo: [NSLocalizedDescriptionKey: "The operation you are trying to perform is not permitted."])
```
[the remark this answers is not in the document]
esreli: You seem to like this pattern. It's nice, sure!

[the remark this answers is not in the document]
philium: I do! I like how it affords using implicit member expressions, although that isn't of much help in this case. Seeing `InvalidOperationError` and the like used at the call sight, I kept thinking, *is that a type?* because they begin with an uppercase letter.

Just thinking about it a little more, since `NSError` is a reference type, it might make sense to make it a computed property rather than a stored property so it doesn't hang around forever. That being said, if it is practically never accessed, it would practically never be created, as static variables are initialized lazily.

data-collection/data-collection/App Errors/LoadablesError.swift
```diff
@@ -0,0 +1,43 @@
...
+        case .multiLoadableFailure(_):
+            return base + 1
+        }
+    }
+    
+    var errorUserInfo: [String : Any] {
+        switch self {
+        case .multiLoadableFailure(let objectLocalizedDescription, let errors):
+            return [NSLocalizedFailureReasonErrorKey: "One or more loadable objects failed to load. Interrogate the \"MultiLoadablesFailureErrors\" key to learn more.",
+                    NSLocalizedDescriptionKey: "Error loading \(objectLocalizedDescription.localizedLowercase).",
```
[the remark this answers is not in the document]
esreli: I'll take this consideration, though many formatted strings shouldn't be formatted because they are formatted with a string that should not be localized, such as a table name.

data-collection/data-collection/Rich Popup/Related Records/Relationship.swift
```diff
@@ -35,44 +35,68 @@ class RelatedRecordsManager {
...
             
             guard error == nil else {
-                completion(nil, error!)
+                self.loadDidFinishWithError(error!)
                 return
             }
             
             guard let popups = popupsResults else {
-                completion(nil, FeatureTableError.queryResultsMissingPopups)
+                self.loadDidFinishWithError(FeatureTableError.queryResultsMissingPopups)
```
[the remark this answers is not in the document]
esreli: I'll swap out for `NSError.unknown`

esreli:
@philium i've made some changes re: your review.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/RichPopup.storyboard
philium: The title is not fixed if you use the view controller to maintain editing state. From the documentation:

> If one of the custom views of the navigationItem property is set to the returned object, the associated navigation bar displays an Edit button if isEditing is false and a Done button if isEditing is true. The default button action invokes the setEditing(_:animated:) method.

It basically does what you are doing manually. It displays Done when the view controller is in an editing state and Edit otherwise. See the documentation for `UIViewController. setEditing(_:animated:)`.

esreli: Ah, I understand now. I thought you were asking if I had considered using the system item Edit button:

`<barButtonItem key="rightBarButtonItem" systemItem="edit" id="Wnf-HI-Nv9">`

No, I haven't considered this. I haven't because I don't intend to use the UITableView editing API. This API is designed to delete, insert and re-order table cells (and their underlying values, of course).

What benefit do I get from using `UITableViewController.setEditing(_:animated:)` to drive the title of the bar button item when I don't intend to use the editing API anyway? It seems like more of a hassle writing that supporting code for a feature that would be written back out.

esreli: I remember implementing this suggestion after you made the suggestion in a previous PR (I think it was the first review you offered of this app). On the surface it seemed like a good one, but when I began implementing it, I was immediately reminded that I was entering the world of the `UITableView` editing API. 

For instance, despite specifying every cell editing style `.none`

```swift
override func tableView(_ tableView: UITableView, editingStyleForRowAt indexPath: IndexPath) -> UITableViewCell.EditingStyle {
    return .none
}
```

Each table view cell adjusts for an editing accessory view, despite specifying the accessory view nil.

screen shot 2018-12-03 at 1 09 09 pm

If I were to implement this suggestion, I would have to ensure the content view isn't adjusted for a missing editing accessory view. Do you know how I might do that? Is it worth it, considering I don't intend to use the API whatsoever?

philium: > Do you know how I might do that?

Check out `UITableViewDelegate.tableView(_:shouldIndentWhileEditingRowAt:)`.

> Is it worth it, considering I don't intend to use the API whatsoever?

That's up to you. I think it is worth exploring. If you get it working, then you can decide if it's worth keeping. I personally would prefer to use the view controller editing API since it is there, even though it means customizing the table view editing behavior through provided APIs, rather than re-implementing some of the editing behavior myself.

esreli: I've filed an issue to take care of this another time.

https://github.com/Esri/data-collection-ios/issues/162

PR #168, opened 2019-03-14
Rich popup view(s) n attributes configurable

esreli:


data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupDetailsViewController/RichPopupDetailsViewController.swift
```diff
@@ -23,6 +23,11 @@ protocol RichPopupDetailsViewControllerDelegate: AnyObject {
 
 class RichPopupDetailsViewController: UITableViewController {
     
+    struct RelatedRecordsConfiguration {
```
philium: I would suggest turning these structs into enums so that they cannot be instantiated.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupDetailsViewController/RichPopupDetailsViewController.swift
```diff
@@ -23,6 +23,11 @@ protocol RichPopupDetailsViewControllerDelegate: AnyObject {
 
 class RichPopupDetailsViewController: UITableViewController {
     
+    enum RelatedRecordsConfiguration: Int {
+        case maxOneToMany = 3
```
philium: FYI: You could have left these as static constants inside the enum. It is perfectly fine to have an enum with no cases. That way you wouldn't need to specify `.rawValue`.

PR #170, opened 2019-03-20
sets minimum font scale for attachment names

esreli:
Before:
_(note the ellipsis)_
Screen Shot 2019-03-20 at 1 43 38 PM

After:
_(note the reduced font size)_
Screen Shot 2019-03-20 at 1 42 55 PM


mikewilburn:
I think this is an excellent solution. Here's another before/after look at the effect of this PR:

image image
image

philium:
As @mikewilburn's screenshots show, it is still possible for names to be truncated. That's fine, but you might consider truncating middle rather than tail. If you want the full name to be visible no matter what, you could let the name label grow to multiple lines.

mikewilburn:
I still think this is a reasonable enough solution that will accommodate for a vast majority of the cases out there.

I also noticed that before the image/file itself will load, the view will display the filename instead:

image

PR #175, opened 2019-04-01
Appends popup image attachment file extension and ensures filename is not empty.

esreli:


data-collection/data-collection/Rich Popup/Attachments/RichPopupStagedPhotoAttachment.swift
```diff
@@ -27,13 +27,30 @@ class RichPopupStagedPhotoAttachment: RichPopupStagedAttachment {
     
     let image: UIImage
     
+    var nameAsJPEG: String? {
+        
+        guard var jpegName = name?.trimmingCharacters(in: .whitespacesAndNewlines) else { return nil }
+        
+        let fileExtension = NSString(string: jpegName).pathExtension
```
philium: I would suggesting bridging `jpegName` so as not to create a new `NSString` object.

```suggestion
        let fileExtension = (jpegName as NSString).pathExtension
```

data-collection/data-collection/Rich Popup/Attachments/RichPopupStagedPhotoAttachment.swift
```diff
@@ -27,13 +27,30 @@ class RichPopupStagedPhotoAttachment: RichPopupStagedAttachment {
...
+        
+        if jpegName.isEmpty {
+            jpegName.append("Image")
+        }
+        
+        if fileExtension != "jpeg" && fileExtension != "jpg" {
```
philium: Another way of doing this is using a set:

```swift
let possibleExtensions: Set<String> = ["jpeg", "jpg"]
if !possibleExtensions.contains(fileExtension) {
    jpegName.append(".jpeg")
}
```

Issue #174, opened 2019-04-01
Resulting attachment size differs from selected PREFERRED SIZE


mikewilburn:
When attaching a new image, the `PREFERRED SIZE` menu permits downsampling the image size to save file space and allow for faster sync times:

![image](https://user-images.githubusercontent.com/25207711/55353151-6db60e80-5477-11e9-8de8-eb9a7130e72e.png)

However, this size appears to be getting ignored and instead honoring 3x scale factor.

For instance, when specifying a size of `Medium (480x640)`, the resulting image size being uploaded is actually `1440x1920`.

mikewilburn:
In the latest version of Collector for iOS, I confirmed that their selected sizes are consistent with what's listed in the iOS API ref [here](https://developers.arcgis.com/ios/latest/api-reference/_a_g_s_enumerations_8h.html#acbeb74606c1a71ec892616b77953fc92).

A size 'Large' does upload an image of size of `960x1280`, however they've removed these discrete size definitions from the corresponding settings menu:
![image](https://user-images.githubusercontent.com/25207711/55353355-ee750a80-5477-11e9-97c5-909ee3765038.png)

esreli:
The SDK downsamples images respecting the main screen's scale. There are a few options available to us to better express to our users what is happening under the hood of the SDK:

1. Specify points in the preferred size options table.
2. Write a multiplier for the preferred size options relative to screen scale, outputting in the preferred size options table the actual down-sample size in pixels.
3. Down-sample the image ourself. (This should be considered last.)

Thoughts? @mikewilburn

esreli:
Adding label `Type - Enhancement` and removing `Type - Bug` because this behavior is as designed. This does not deny the need to address this concern and merits an examination of how this can be more easily understood.

mikewilburn:
My preference would be for number **2** above.

I think the attached file size should be consistent with what's listed in the preferred size options table and the API documentation. This would achieve that.

As well, I don't think an end user should have to consider a scale factor when attaching images. It should be handled behind the scenes, either by the app's logic, or by the API itself (more ideal).

esreli:
This changes has been implemented. Please confirm this issue can be closed.

mikewilburn:
Changes confirmed in branch `v.next`.

On an iPhone X (which has a 3x scale factor), the new preferred sizes are:
![image](https://user-images.githubusercontent.com/25207711/55595808-6488a400-56fa-11e9-8818-237bdbb53b87.png)

PR #179, opened 2019-04-04
Supports image attachment preferred content size with device screen scale

esreli:
In editing an image attachment preferred content size, up-scales output string with the device's main screen scale. This more accurately reflects the image scaling performed by the SDK.

### DIFF
--- a/data-collection/data-collection/Extensions/ArcGIS/AGSPopupAttachmentSize/AGSPopupAttachmentSize+String.swift
+++ b/data-collection/data-collection/Extensions/ArcGIS/AGSPopupAttachmentSize/AGSPopupAttachmentSize+String.swift
@@ -32,20 +32,19 @@ extension AGSPopupAttachmentSize {
     }
     
     var actualSizeScaled: CGSize {
-        let size = actualSize
         let scale = UIScreen.main.scale
-        return CGSize(width: size.width * scale, height: size.height * scale)
+        return actualSize.applying(.init(scaleX: scale, y: scale))
     }
     
     var actualSizeTitle: String {
-        return titleFor(size: actualSize)
+        return title(forSize: actualSize)
     }
     
     var actualSizeScaledTitle: String {
-        return titleFor(size: actualSizeScaled)
+        return title(forSize: actualSizeScaled)
     }
     
-    private func titleFor(size: CGSize) -> String {
+    private func title(forSize size: CGSize) -> String {
         
         let width = Int(size.width)
         let height = Int(size.height)
--- a/data-collection/data-collection/Extensions/ArcGIS/AGSPopupAttachmentSize/AGSPopupAttachmentSize+String.swift
+++ b/data-collection/data-collection/Extensions/ArcGIS/AGSPopupAttachmentSize/AGSPopupAttachmentSize+String.swift
@@ -31,10 +31,25 @@ extension AGSPopupAttachmentSize {
         }
     }
     
-    var asTitle: String {
+    var actualSizeScaled: CGSize {
         let size = actualSize
+        let scale = UIScreen.main.scale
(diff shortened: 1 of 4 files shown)


esreli:
Related issue: https://github.com/Esri/data-collection-ios/issues/174

data-collection/data-collection/Extensions/ArcGIS/AGSPopupAttachmentSize/AGSPopupAttachmentSize+String.swift
```diff
@@ -31,10 +31,25 @@ extension AGSPopupAttachmentSize {
...
+    
+    var actualSizeTitle: String {
+        return titleFor(size: actualSize)
+    }
+    
+    var actualSizeScaledTitle: String {
+        return titleFor(size: actualSizeScaled)
+    }
+    
+    private func titleFor(size: CGSize) -> String {
```
philium: According to the Swift API Design Guidelines, "When the first argument forms part of a prepositional phrase…the argument label should normally begin at the preposition." So `title(forSize:)` might be a more appropriate title for this method.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupEditStagedPhotoAttachmentViewController/RichPopupEditStagedAttachmentViewController.swift
```diff
@@ -185,4 +185,16 @@ class RichPopupEditStagedAttachmentViewController: UITableViewController, UIText
             return 1
         }
     }
+    
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        
+        if indexPath.section == .nameSection {
```
philium: You might consider condensing this code a little bit by retrieving the cell in only one place:

```swift
let cell = super.tableView(tableView, cellForRowAt: indexPath)
if indexPath.section == .preferredSizeSection {
    cell.textLabel?.text = AGSPopupAttachmentSize.tableViewMap[indexPath.row].actualSizeScaledTitle
}
return cell
```

data-collection/data-collection/Extensions/ArcGIS/AGSPopupAttachmentSize/AGSPopupAttachmentSize+String.swift
```diff
@@ -31,10 +31,25 @@ extension AGSPopupAttachmentSize {
         }
     }
     
-    var asTitle: String {
+    var actualSizeScaled: CGSize {
         let size = actualSize
+        let scale = UIScreen.main.scale
+        return CGSize(width: size.width * scale, height: size.height * scale)
```
philium: Another way of doing this is to apply a transform:

```swift
return size.applying(.init(scaleX: scale, y: scale))
```

mikewilburn:
The changes introduced by this PR look good.

I confirmed on `eli.preferred-size` the following behaviors:
1. The `PREFERRED SIZE` list accurately reflects the final resolutions to which the attached images will be downsampled. Screen scaling factors are now honored).
2. Both portrait and landscape orientations are respected. 3:4 is the same as 4:3. Long edges are treated equivalently.
 i.e. a landscape image selected to be attached a resolution of 480x640 will have an uploaded resolution of 640x480, not 480x360.
3. Images that do not have the 3:4 width-to-height ratio will be scaled to ensure they fit _entirely_ into the selected resolution.
 i.e. a 16:9 image selected to be attached at a resolution of 720x960 will have an uploaded resolution of 960x540, not 720x1280.

PR #183, opened 2019-04-05
Migrates project to compile for swift 5

esreli:


philium:
I didn't think of it at the time, but you'll want to update the README. It currently lists Xcode 9 as a requirement; that should be bumped to Xcode 10.2 for Swift 5.

PR #185, opened 2019-04-10
Adds entitlements build phase run script

esreli:


nixta:
Should this be patched onto the end of the existing script, or added as a separate script? I suspect the latter option would be preferred.

esreli:
>  I suspect the latter option would be preferred.

I don't mind creating a separate script. I'm curious though why you think this is preferred?

nixta:
They do different things. I also don't know if they should be doing them at the same position in the build phase steps (but probably it doesn't matter).

But mostly, having it as a separate phase step relates it more cleanly to the SDK guide instructions.

esreli:
@nixta i've incorporated your suggestion. Have another look?

data-collection/data-collection.xcodeproj/project.pbxproj
```diff
@@ -1356,6 +1357,23 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
+		94476EB3226A97F400EED38E /* ShellScript */ = {
```
philium: Since there's already a Run Script phase, what would you think of giving this a different name, maybe something like "Strip Frameworks"?

data-collection/data-collection.xcodeproj/project.pbxproj
```diff
@@ -1221,6 +1221,7 @@
 				94FA8B1C207BF2CD00972C75 /* Frameworks */,
 				94FA8B1D207BF2CD00972C75 /* Resources */,
 				94AEFD2A217F8F69004339B9 /* ShellScript */,
+				94476EB3226A97F400EED38E /* ShellScript */,
```
philium: Per the [guide](https://developers.arcgis.com/ios/latest/swift/guide/install.htm), this should come after the `Embed Frameworks` phase:

> Ensure that the `Run Script` phase follows the **Embed Frameworks** phase in sequence otherwise the script may not run correctly. This may happen, for instance, if you remove and re-add the ArcGIS framework after creating the Run Script phase.

PR #189, opened 2019-05-30
Updates copy to reflect the ArcGIS style guide concerning usage of "sign in".

esreli:
TLDR;
"Sign in" is the verb, "Login" is the noun.

### DIFF
--- a/data-collection/data-collection/Extensions/ArcGIS/AGSPortal/AGSPortal+CachedCredentials.swift
+++ b/data-collection/data-collection/Extensions/ArcGIS/AGSPortal/AGSPortal+CachedCredentials.swift
@@ -19,9 +19,9 @@ extension AGSPortal {
     
     /// Returns an `AGSPortal` that is signed in if cached credentials allow.
     ///
-    /// The credential cache will have at most one valid cached credential. It is used to sign-in to the portal if possible.
+    /// The credential cache will have at most one valid cached credential. It is used to sign in to the portal if possible.
     ///
-    /// If cached credentials do not automatically sign-in to the portal, then connect to
+    /// If cached credentials do not automatically sign in to the portal, then connect to
     /// the portal anonymously (i.e. not signed in).
     ///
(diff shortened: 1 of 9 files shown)


data-collection/data-collection/Extensions/ArcGIS/AGSPortal/AGSPortal+CachedCredentials.swift
```diff
@@ -17,27 +17,27 @@ import ArcGIS
...
-        // then in the portal.load() callback below we will find ourselves logged in to the portal.
+        // then in the portal.load() callback below we will find ourselves signed-in to the portal.
         let newPortal = AGSPortal(url: portalURL, loginRequired: true)
         
-        // We'll temporarily disable prompting the user to log in in case the cached credentials are not suitable to log us in.
-        // I.e. if the cached credentials aren't good enough to find ourselves logged in to the portal/ArcGIS Online, then just
-        // accept it and don't prompt us to log in, resulting in a Portal being accessed anonymously.
+        // We'll temporarily disable prompting the user to sign-in in case the cached credentials are not suitable to sign us in.
+        // I.e. if the cached credentials aren't good enough to find ourselves signed in to the portal/ArcGIS Online, then just
+        // accept it and don't prompt us to sign in, resulting in a Portal being accessed anonymously.
```
nixta: I'm a bit confused by the `sign in` vs `sign-in` here. For example, in line 22 above, I wonder if it really needs to be `sign-in` or on line 35 that it needs to be `signed-in` vs `signed in`.

But this then appears inconsistent. I haven't checked with the style guide, so I'm not myself entirely sure what the right answer is but there appears to be a little inconsistency here.

esreli: That's my fault! As a verb, we use "sign in", as an adjective we use "sign-in", as a noun we use "login".

PR #193, opened 2019-07-25
Use Result for reverse geocode completion parameter

philium:


esreli:
We have no immediate slated future release schedule for data-collection, please modify the base branch to `master`.

Issue #197, opened 2019-08-07
Update AlamoFire NetworkReachabilityManager


esreli:

esreli:
https://github.com/Alamofire/Alamofire/blob/master/Source/NetworkReachabilityManager.swift

PR #199, opened 2019-08-07
Updates app with latest Alamofire NetworkReachabilityManager

esreli:
https://github.com/Alamofire/Alamofire/blob/master/Source/NetworkReachabilityManager.swift


esreli:
https://github.com/Esri/data-collection-ios/issues/197

Issue #195, opened 2019-08-06
Migrate app to Swift 5.1


esreli:

mhdostal:
We're already updated to Swift 5.1.

PR #200, opened 2019-08-07
Update app to swift 5.1

esreli:
Now triggers fatal error for unsupported enum cases.

### DIFF
--- a/data-collection/data-collection/AppLocation.swift
+++ b/data-collection/data-collection/AppLocation.swift
@@ -66,6 +66,8 @@ extension CLAuthorizationStatus: CustomStringConvertible {
             return "Not Determined"
         case .restricted:
             return "Restricted"
+        @unknown default:
+            fatalError("Unsupported case \(self).")
         }
     }
 }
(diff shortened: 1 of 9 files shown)


esreli:
https://github.com/Esri/data-collection-ios/issues/195

esreli:
Closing. Silly mistake. I was under the _wrong_ impression that [non exhaustive enums ](https://github.com/apple/swift-evolution/blob/master/proposals/0192-non-exhaustive-enums.md)was to be introduced in Swift 5.1 but is already introduced in swift 5.

https://swift.org/migration-guide-swift5/

> n Swift 5 mode, switches over enums, declared in Objective-C or coming from system frameworks, are required to handle “unknown cases”, i.e. cases that might be added in the future, or that may be defined “privately” in an Objective-C implementation file. (Formally, Objective-C allows storing any value in an enum as long as it fits in the underlying type.) These “unknown cases” can be handled by using the new @unknown defaultcase, which still provides warnings if any known cases are omitted from the switch. They can also be handled using a normal default.

PR #207, opened 2019-08-08
Consolidate creation of OK and Cancel actions

philium:
As a result, the action titles and styles only have to be specified in one place.

### DIFF
--- a/data-collection/data-collection/Extensions/UIKit/UIAlertController/UIAlertController+Alert.swift
+++ b/data-collection/data-collection/Extensions/UIKit/UIAlertController/UIAlertController+Alert.swift
@@ -18,20 +18,19 @@ extension UIAlertController {
     
     /// Build a single button `UIAlertController` with a message.
     ///
-    /// You may optionally provide additional parameters including the title, the action button title and an action closure.
+    /// You may optionally provide additional parameters including the title and an action closure.
     ///
     /// - Parameters:
     ///   - title: The title of the alert (optional).
     ///   - message: The message body of the alert.
-    ///   - actionTitle: The title of the alert's dismiss button (optional), the default value is `"OK"`.
     ///   - action: A closure triggered upon tapping the action button (optional).
     ///
     /// - Returns: A newly created and configured single button `UIAlertController` of style `.alert`.
-    
-    static func simpleAlert(title:String? = nil, message:String, actionTitle:String = "OK", action:((UIAlertAction)->Void)? = nil) -> UIAlertController {
+    static func simpleAlert(title: String? = nil, message: String, action: ((UIAlertAction) -> Void)? = nil) -> UIAlertController {
         let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
-        let action = UIAlertAction(title: actionTitle, style: .default, handler: action)
+        let action = UIAlertAction.okay(handler: action)
         alert.addAction(action)
+        alert.preferredAction = action
         return alert
     }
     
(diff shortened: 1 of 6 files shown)


data-collection/data-collection/Utilities/ImagePickerPermissions.swift
```diff
@@ -148,6 +145,10 @@ public struct ImagePickerPermissions {
                         alert.addAction(settingsAction)
                     }
                     
+                    let okAction = UIAlertAction.okay()
```
philium: I think the correct placement for this button is on the right:

> In general, buttons people are most likely to tap should be on the right.

The user has already denied access, so it seems like the most likely action would be to continue. It also matches system alerts, like the following:

image

data-collection/data-collection/Utilities/ImagePickerPermissions.swift
```diff
@@ -136,9 +136,6 @@ public struct ImagePickerPermissions {
                     
                     let alert = UIAlertController(title: "Not Authorized", message: "Go to Settings and authorize the use of the \(permission.title).", preferredStyle: .alert)
                     
-                    let okAction = UIAlertAction(title: "OK", style: .cancel)
```
philium: I believe the correct style for this button should be `default`, based on the title:

> A button that cancels an alert’s action should always be labeled Cancel.

data-collection/data-collection/Extensions/UIKit/UIAlertController/UIAlertController+Alert.swift
```diff
@@ -18,20 +18,19 @@ extension UIAlertController {
...
     ///
     /// - Returns: A newly created and configured single button `UIAlertController` of style `.alert`.
-    
-    static func simpleAlert(title:String? = nil, message:String, actionTitle:String = "OK", action:((UIAlertAction)->Void)? = nil) -> UIAlertController {
+    static func simpleAlert(title: String? = nil, message: String, action: ((UIAlertAction) -> Void)? = nil) -> UIAlertController {
         let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
-        let action = UIAlertAction(title: actionTitle, style: .default, handler: action)
+        let action = UIAlertAction.okay(handler: action)
         alert.addAction(action)
+        alert.preferredAction = action
```
esreli: If there is only one action, is this necessary?

data-collection/data-collection/Extensions/UIKit/UIAlertController/UIAlertController+Alert.swift
```diff
@@ -45,17 +44,14 @@ extension UIAlertController {
...
     ///   - cancel: A closure triggered upon tapping the cancel button (optional).
     ///
     /// - Returns: A newly created and configured two-button button `UIAlertController` of style `.alert`.
-    
-    static func multiAlert(title:String? = nil, message:String, actionTitle:String, action:((UIAlertAction)->Void)? = nil, isDestructive: Bool = false, cancelTitle:String = "Cancel", cancel:((UIAlertAction)->Void)? = nil) -> UIAlertController {
+    static func multiAlert(title: String? = nil, message: String, actionTitle: String, action: ((UIAlertAction) -> Void)? = nil, isDestructive: Bool = false, cancel cancelHandler: ((UIAlertAction) -> Void)? = nil) -> UIAlertController {
         let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
         let action = UIAlertAction(title: actionTitle, style: isDestructive ? .destructive : .default, handler: action)
-        let cancel = UIAlertAction(title: cancelTitle, style: .cancel, handler: cancel)
         alert.addAction(action)
```
esreli: It feels as though preferred action belongs down here, no?

data-collection/data-collection/Utilities/ImagePickerPermissions.swift
esreli: Can you name the variable `okayAction`?
```suggestion
                    let okayAction = UIAlertAction.okay()
```

data-collection/data-collection/Utilities/ImagePickerPermissions.swift
```diff
@@ -148,6 +145,10 @@ public struct ImagePickerPermissions {
                         alert.addAction(settingsAction)
                     }
                     
+                    let okAction = UIAlertAction.okay()
+                    alert.addAction(okAction)
+                    alert.preferredAction = okAction
```
esreli: Having the user authorize the use of location services feels to me the preferred action, no?

data-collection/data-collection/Extensions/UIKit/UIAlertController/UIAlertController+Alert.swift
philium: It isn't necessary, but last I checked, it matches the look of Apple's single button alerts.

data-collection/data-collection/Extensions/UIKit/UIAlertController/UIAlertController+Alert.swift
philium: Yes, I think it does make sense for the non-cancel action to be the preferred action. I'll add it.

data-collection/data-collection/Utilities/ImagePickerPermissions.swift
philium: Yep. I'll change it.

data-collection/data-collection/Utilities/ImagePickerPermissions.swift
philium: It is certainly the action we'd want them to take so that we can do whatever we do with their location. But the HIG says "In general, buttons people are most likely to tap should be on the right." Since the user has already explicitly denied the app access to their location, is it likely that they will then go grant that access? I'm not so sure. That being said, the order of the buttons isn't that big of a deal. I can move the OK button back to the left if you'd like.

PR #210, opened 2019-09-24
Rename example apps

jf990:
one outdated reference


esreli:
Because this change is so minute and resolves an issue in the README and not source code, I do not deem this change worth of a new patch release.

Issue #188, opened 2019-05-21
App doesn't validate attachment file name


codergrl:
When renaming a new attachment, the app should validate so the user cannot enter invalid characters like spaces, backslashes, etc. Currently the app allows the user to enter invalid characters, and then it fails to save the attachment throwing a null pointer exception. 

Here's the regex used in the dotnet app to validate: `^[^\\\/:*?""<>|\s]+$`

nCastle1:
It looks like the editing is a separate issue (affects master as well) so I'll open a separate issue.

mhdostal:
> I tried the branch with the fix. I noticed that it still allowed me to enter `]`, but it was saved as `_`.

It's fine with me iff there are characters that the system can handle but get translated; it just can't fail to save the attachment.

mhdostal:
Verified fixed in the v1.1.2 release.

PR #212, opened 2019-10-14
Add check for illegal attachment name characters.

mhdostal:
#188 

The regular expression used to find illegal characters:  ^[\\^\\\\:\\/*?\"<>|]+$

So Essentially:  ^\:/*?"<>|

Note that for iOS, whitespace is a valid character.


data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupEditStagedPhotoAttachmentViewController/RichPopupEditStagedAttachmentViewController.swift
```diff
@@ -106,6 +106,18 @@ class RichPopupEditStagedAttachmentViewController: UITableViewController, UIText
         return true
     }
     
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
```
philium: Correct me if I'm wrong, but wouldn't this be an equivalent implementation for this method?

```swift
return string.rangeOfCharacter(from: CharacterSet(charactersIn: #"^\:/*?"<>|"#)) == nil
```

If so, I think it would be more efficient as it should bail as soon as it encounters an invalid character.

mhdostal: Yes, but it's not as cool as regular expressions.  I was overthinking the problem, thanks.

Issue #209, opened 2019-09-18
Crash when trying to add image as attachment (iPad)


mhdostal:
When attempting to add an attachment to a new tree, tapping the "Add Attachments" row in the "Attachments" view crashes the app.  The error shown in the log is:

```
2019-09-18 11:17:26.072695-0500 data-collection[3197:356042] *** Terminating app due to uncaught exception 'NSGenericException', reason: 'Your application has presented a UIAlertController (<UIAlertController: 0x11d272000>) of style UIAlertControllerStyleActionSheet from data_collection.AppContextAwareNavigationController (<data_collection.AppContextAwareNavigationController: 0x11d95ea00>). The modalPresentationStyle of a UIAlertController with this style is UIModalPresentationPopover. You must provide location information for this popover through the alert controller's popoverPresentationController. You must provide either a sourceView and sourceRect or a barButtonItem.  If this information is not known when you present the alert controller, you may provide it in the UIPopoverPresentationControllerDelegate method -prepareForPopoverPresentation.'
``` 

The cause is the UIAlertController needs a rectangle to display the action sheet from on an iPad.  This can happen either by the controller's `popoverPresentationController?.barButtonItem` or a combination of `popoverPresentationController?.sourceView` and `popoverPresentationController?.sourceRect` properties.

The solution is to provide a source Rect that the popover would display from (the cell's frame maybe?) or to switch to a different type of alert VC (maybe).

This is iPad-only; does not crash on an iPhone 8.

iPad Pro 10.5"
iOS 12.4.1

mhdostal:
The same crash will occur when attempting to add a new feature and there is more than one feature layer in the map.  When there is more than one FL, an AlertController is displayed allowing the user to select which FL to add the feature two.  The default configuration in the app has only one FL, so this crash would not happen in that case.

mhdostal:
There are two separate instances of the crash with the `UIAlertViewController`: image source selection and Feature Layer selection.

The first, image source selection can be solved by changing the type of the `UIAlertController` to `.alert` from `.actionSheet`.  There is no barButtonItem in this case as it's triggered by tapping on the `AddAttachment` row in a table.  Also, the display of the action sheet is many layers removed, code-wise, from the table view so that passing in a source view and rect is not practical.  So in this case, `.alert` is the best bet.  It will look like this:

![image](https://user-images.githubusercontent.com/3998072/65259919-1cecad80-dacb-11e9-9cb4-fa2416235a25.png)

[...]

PR #217, opened 2019-10-21
Mhd/i os13 change

mhdostal:
This reverts the change to using an `.alert` when selecting the image source type for a new attachment on iOS 13.  A fix was installed in a separate PR to use the `.alert` style of `UIAlertController` because using `.actionSheet` crashed.  This crash has been resolved in iOS 13 and since `.actionSheet` is the preferred style to use, use it for iOS 13, but keep `.alert` for pre-13 devices.


mhdostal:
https://github.com/Esri/data-collection-ios/issues/209

data-collection/data-collection/Utilities/ImagePickerPermissions.swift
```diff
@@ -174,7 +174,13 @@ public struct ImagePickerPermissions {
         
         let viewController = delegate.imagePickerPermissionsRequestsPresentingViewController()
         
-        let alertController = UIAlertController(title: nil, message: nil, preferredStyle: .alert)
+        var style = UIAlertController.Style.alert
```
esreli: This is a nice touch. You might consider this alternative that is functionally equivalent but avoids setting `style` twice (under a certain circumstance).

```swift
let style: UIAlertController.Style

if #available(iOS 13.0, *) {
    style = .actionSheet
}
else {
    style = .alert
}
```

data-collection/data-collection/Utilities/ImagePickerPermissions.swift
```diff
@@ -174,7 +174,16 @@ public struct ImagePickerPermissions {
         
         let viewController = delegate.imagePickerPermissionsRequestsPresentingViewController()
         
-        let alertController = UIAlertController(title: nil, message: nil, preferredStyle: .alert)
+        var style: UIAlertController.Style
```
esreli: This is very nit-picky but I believe this can be a `let`, no? Would be a tiny compiled optimization.

mhdostal: I am setting the style based on the availability of iOS 13 or not.

PR #228, opened 2020-03-11
V.next

mhdostal:
This merges v.next into master in preparation for the next release.


mhdostal:
@esreli - got it, thanks for the reminder!

PR #230, opened 2020-04-21
Mhd/layer contents integration

mhdostal:
Included in this PR:

- LayerContents component based on [this](https://github.com/ArcGIS/open-source-apps/blob/master/app/data-collection/components/layer-contents/Spec.md) design.  Note that the title and title icon for both the Layers and Bookmarks view is under design review and not included here.

- Inclusion of an "Extras" button in the right side of the app's title bar which displays an AlertViewController allowing the user to select either Layers or Bookmarks.

- Integration of the LayerContents component in DataCollection.

- Integration of the Bookmarks component from the Toolkit (and inclusion of the Toolkit as a submodule).

Extras alert:
Extras

Layers:
Layers

Bookmarks:
Bookmarks


data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,113 @@
...
+                image = #imageLiteral(resourceName: "legend")
+            case bookmarksExtra:
+                image = #imageLiteral(resourceName: "bookmark")
+            default:
+                break
+            }
+            
+            extraAction.setValue(image, forKey: "image")
+            action.addAction(extraAction)
+        }
```
philium, line 55: It seemed like the `for` loop was a convenience…until I saw the two switch statements. Now I'm wondering whether the code would be simpler without the for loop.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,113 @@
...
+            switch extra {
+            case layersExtra:
+                image = #imageLiteral(resourceName: "legend")
+            case bookmarksExtra:
+                image = #imageLiteral(resourceName: "bookmark")
+            default:
+                break
+            }
+            
+            extraAction.setValue(image, forKey: "image")
```
philium, line 53: I really don't recommend doing this. I know that it works now, but there's no guarantee it will continue to work in the future. As much as I would love for Apple to expose a way to set an image on an alert action, they haven't, and I can only assume there's a good reason for that. Apple is [pretty clear](https://developer.apple.com/documentation/uikit/windows_and_screens/getting_the_user_s_attention_with_alerts_and_action_sheets) about what alerts and action sheets should be used for ("important information…important choice"). I've been guilty of violating that and using an action sheet out of convenience, but I don't recommend it. If at all possible, implement your own view for this.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,113 @@
...
+        if layerContentsViewController == nil {
+            let dataSource = DataSource(geoView: mapView)
+            layerContentsViewController = TableOfContents(dataSource)
+            
+            // Add a done button.
+            let doneButton = UIBarButtonItem(barButtonSystemItem: .done, target: self, action: #selector(done))
+            layerContentsViewController?.navigationItem.leftBarButtonItem = doneButton
+        }
+        
+        if let layerContentsVC = layerContentsViewController {
```
philium, line 73: You could make the second `if` statement unnecessary by using an `if…else` statement:

```swift
let layerContentsVC: LayerContentsViewController
if let existingViewController = layerContentsViewController {
    layerContentsVC = existingViewController
} else {
    // create and configure the view controller
    layerContentsVC = ...
}
```

Same in `showBookmarks(_:)`.

data-collection/data-collection/View Controllers/App Container/Base.lproj/AppContainerViewController.storyboard
```diff
@@ -85,6 +83,11 @@
                             </connections>
                         </barButtonItem>
                         <rightBarButtonItems>
+                            <barButtonItem title="Item" image="ellipsis" catalog="system" id="4IE-fC-WYU" userLabel="Extras Item">
```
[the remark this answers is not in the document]
mhdostal: That is odd that it's the default;  I made the change to `plain` on all of the bar button items.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/DataSource.swift
```diff
@@ -0,0 +1,126 @@
...
+/// The data source is used to represent an array of `AGSLayerContent` for use in a variety of
+/// implementations. It is initialized with either an array of `AGSLayerContent`
+/// or an `AGSGeoView` from whose `AGSMap` or `AGSScene` the  operational and
+/// base map layers (`AGSLayerContent`) are extracted.
+/// - Since: 100.8.0
+public class DataSource: NSObject {
+    /// Returns a `DataSource` initialized with the given `AGSLayerContent` array..
+    /// - Parameter layers: The array of `AGSLayerContent`.
+    /// - Since: 100.8.0
+    public init(layers: [AGSLayerContent]) {
```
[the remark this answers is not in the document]
mhdostal: I'm cool with that.  @esreli - You OK with this suggestion?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/DataSource.swift
```diff
@@ -0,0 +1,126 @@
...
+
+import UIKit
+import ArcGIS
+
+/// The data source is used to represent an array of `AGSLayerContent` for use in a variety of
+/// implementations. It is initialized with either an array of `AGSLayerContent`
+/// or an `AGSGeoView` from whose `AGSMap` or `AGSScene` the  operational and
+/// base map layers (`AGSLayerContent`) are extracted.
+/// - Since: 100.8.0
+public class DataSource: NSObject {
```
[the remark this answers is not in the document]
mhdostal: We did at one point, but not sure why we didn't make the change;  @esreli ?

data-collection/data-collection/View Controllers/App Container/Base.lproj/AppContainerViewController.storyboard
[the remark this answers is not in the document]
mhdostal: An oversight on my part; I was going to change it but just forgot about it.  Thanks!  I changed it to the Calcite Ellipsis.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/DataSource.swift
```diff
@@ -0,0 +1,126 @@
...
+        if let geoView = geoView {
+            geoView.layerViewStateChangedHandler = { [weak self] (layer: AGSLayer, layerViewState: AGSLayerViewState) in
+                self?.geoViewDidChange()
+            }
+        }
+    }
+    
+    private func appendBasemap(_ basemap: AGSBasemap?) {
+        guard let basemap = basemap else { return }
+
```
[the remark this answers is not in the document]
mhdostal: I made the change and removed an unnecessary `let mapView = ...`.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.storyboard
```diff
@@ -0,0 +1,199 @@
...
+                                <connections>
+                                    <outlet property="accordionButton" destination="ihQ-ML-iVw" id="eqn-6q-YN5"/>
+                                    <outlet property="accordionButtonWidthConstraint" destination="pxE-W6-fcs" id="RG2-My-PUE"/>
+                                    <outlet property="indentationConstraint" destination="5D5-eq-j5c" id="F4o-tb-d5u"/>
+                                    <outlet property="nameLabel" destination="qCc-Am-7U7" id="Rg1-24-oaQ"/>
+                                    <outlet property="separatorView" destination="QXC-nS-UuU" id="uUX-p9-BrB"/>
+                                    <outlet property="visibilitySwitch" destination="Qjr-f6-CC9" id="49i-PM-IbO"/>
+                                </connections>
+                            </tableViewCell>
+                            <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" selectionStyle="default" indentationWidth="10" reuseIdentifier="SublayerTitle" id="bWm-du-xG7" customClass="LayerCell" customModule="LayerContentsForiPad" customModuleProvider="target">
```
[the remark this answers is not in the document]
mhdostal: The `Name` label uses different text in the `LayerTitle` vs. `SublayerTitle` cells.  The functionality is identical, but I wanted two different cells for customization purposes.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+let indentationConstant: CGFloat = 10.0
+
+/// The protocol you implement to respond to user content accordion changes.
+internal protocol ContentAccordionDelegate: AnyObject {
```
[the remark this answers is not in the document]
mhdostal: Good suggestion.  Done.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+            accordionButtonWidthConstraint.constant = !accordionButton.isHidden ? accordionButton.frame.height : 0.0
+            accordionButton.setImage(LayerContentsTableViewController.accordianImage(content?.accordion ?? .none), for: .normal)
+            visibilitySwitch.isHidden = !(content?.allowToggleVisibility ?? false)
+            visibilitySwitch.isOn = content?.isVisibilityToggleOn ?? false
+            
+            layerIndentationLevel = content?.indentationLevel ?? 0
+        }
+    }
+    
+    var layerIndentationLevel: Int = 0 {
```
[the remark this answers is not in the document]
mhdostal: Yes, and it didn't work for me for some reason.

data-collection/data-collection/View Controllers/App Container/Base.lproj/AppContainerViewController.storyboard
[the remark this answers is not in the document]
mhdostal: I reverted this change because I got the following warning after changing them all to `plain`.

image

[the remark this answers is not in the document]
philium: Fascinating! That isn't at all confusing. Good to know!

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.storyboard
```diff
@@ -0,0 +1,199 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
mhdostal: - I'm getting two constraint warnings.  The first has to do with displaying the `UIAlertController` as an `.actionSheet`.  I think that one is a bug on Apple's end.  I get the same message when doing the same thing in the AR Toolkit and it is present in the iOS Samples app as well (in the `DownloadPreplannedMapViewController.removeDownloadsTapped` method).  I found [this](https://stackoverflow.com/questions/55653187/swift-default-alertviewcontroller-breaking-constraints) StackOverflow post.

The second one is: `Warning once only: Detected a case where constraints ambiguously suggest a height of zero for a table view cell's content view.`  I suspect it's because the cell is being displayed before the data is loaded, but I haven't looked into it.

- The Done/Cancel button thing is something that is under design review.

- The rows shouldn't be selectable; I'll fix that.

- fixed the cell spacing.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+    @IBOutlet var nameLabel: UILabel!
+    @IBOutlet var legendImageView: UIImageView!
+    @IBOutlet var activityIndicatorView: UIActivityIndicatorView!
+    @IBOutlet var indentationConstraint: NSLayoutConstraint!
+}
+
+/// LayerCell - cell representing either a Layer or LayerContent (i.e., a sublayer).
+/// The difference is in which table view cell prototype is used for the cell.
+class LayerCell: UITableViewCell {
+    var content: Content? {
```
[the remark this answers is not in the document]
mhdostal: I removed the `contents` property from the TableViewCells and put all that logic into the TableViewDataSource methods.  I needed to redo the LayerCellDelegate as well, but I think it's better for it.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+    
+    @IBAction func accordionAction(_ sender: Any) {
+        guard let button = sender as? UIButton,
+            let accordian = self.content?.accordion,
+            accordian != AccordionDisplay.none  else { return }
+
+        let newAccordian: AccordionDisplay = (content?.accordion == .expanded) ? .collapsed : .expanded
+        content?.accordion = newAccordian
+        button.setImage(LayerContentsTableViewController.accordianImage(newAccordian), for: .normal)
+        accordionDelegate?.accordionChanged()
```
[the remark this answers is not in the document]
mhdostal: I would prefer to only update what changed, but the issue with the accordion functionality is that expanding/contracting the accordion can add or remove several rows in the table.  The `visibleContent` array is what's currently shown.  That is created from the `contents` array based on whether an item's parents.accordion is collapsed or not.  I could do a diff before and after the change, but then I'd have to match the `contents` to the TableViewCell/indexPath and update the table with that.  If I get a chance I'll look into it.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/DataSource.swift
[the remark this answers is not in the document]
esreli: I think the reason it did not change was that I had thought it would serve well as a member of `LayerContentsTableViewController`. For example:

```swift
class LayerContentsTableViewController {
    class DataSource { 
        /* ... */ 
    }
}
```

data-collection/data-collection/View Controllers/Layer Contents View Controllers/DataSource.swift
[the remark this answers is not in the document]
esreli: Yes!

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,474 @@
...
+    
+    /// Using the DataSource's `layercontents` as a starting point, generate the list of `AGSLayerContent` to include in the table view.
+    private func generateLayerList() {
+        // Remove all saved data.
+        legendInfos.removeAll()
+        symbolSwatches.removeAll()
+        contents.removeAll()
+        displayedLayers.removeAll()
+        
+        guard let layerContents = dataSource?.layerContents else { layerContentsTableViewController?.contents = [Content](); return }
```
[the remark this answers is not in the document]
mhdostal: Rewritten to be more understandable.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
mhdostal: rewrote this bit, please take another look.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
mhdostal: I'm surprised it's not supported because it does add some flair to an otherwise unremarkable interface element.  I know I've seen it in other things we've done, but I can't seem to find that now, so maybe it was deleted there as well.  I'll delete them.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+                    cell.tag = indexPath.hashValue
+                    legendInfoCell.symbolImage = nil
+                    
+                    // We don't have a swatch for the given symbol, so create the swatch.
+                    symbol.createSwatch(completion: { [weak self] (image, _) -> Void in
+                        // Make sure this is the cell we still care about and that it
+                        // wasn't already recycled by the time we get the swatch.
+                        if cell.tag != indexPath.hashValue {
+                            return
+                        }
```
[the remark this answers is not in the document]
mhdostal: I used the `symbol.hashValue` instead of the `indexPath.hashValue`.  That way we know that the cell in question contains the symbol the image is retrieved from.  And I'm getting the `indexPath` for the cell and just updating that row.

mhdostal:
@philium - I'm done making the suggested changes and addressing comments, so please give it another look.

The outstanding issues:

- any potential renaming/moving of the DataSource class
- LayerContentsTableViewController issues:  Done/Cancel button and TableViewCell constraint issue
- Find a strategy to eliminate unnecessary loading of tableView data.

Thanks again for your review, it really helps make the code better.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,480 @@
...
+public enum ConfigurationStyle {
+    // Displays all layers.
+    case allLayers
+    // Only displays layers that are in scale and visible.
+    case visibleLayersAtScale
+}
+
+/// Configuration is an protocol (interface) that drives how to format the layer contents table.
+/// - Since: 100.8.0
+public protocol LayerContentsConfiguration {
```
philium, line 30: Unless I'm missing something, adopting types are identical, which makes me think this might make more sense as a struct than a protocol. Then `TableOfContents` and `Legend` could just be static members of the type:

```swift
struct LayerContentsConfiguration { ... }

extension LayerContentsConfiguration {
    /// The configuration for a table of contents.
    static let tableOfContents = ...
    /// The configuration for a legend.
    static let legend = ...
}
```

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,480 @@
...
+    var showRowSeparator: Bool { get }
+    
+    /// The title of the view.
+    /// - Since: 100.8.0
+    var title: String { get }
+}
+
+/// Describes a `LayerContents` view configured to display a legend.
+/// - Since: 100.8.0
+public class Legend: LayerContentsViewController {
```
philium, line 69: Could this be named `LegendViewController`?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,480 @@
...
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+/// Describes a `LayerContents` view configured to display a table of contents.
+/// - Since: 100.8.0
+public class TableOfContents: LayerContentsViewController {
```
philium, line 82: Could this be named `TableOfContentsViewController`?

That being said, it seems like these subclasses aren't even needed. Incorporating my previous suggestion, wouldn't the following be equivalent?

```swift
LayerContentsViewController(dataSource: dataSource, config: .legend) // Legend(dataSource: dataSource)
LayerContentsViewController(dataSource: dataSource, config: .tableOfContents) // TableOfContents(dataSource: dataSource)
```

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+                    // We don't have a swatch for the given symbol, so create the swatch.
+                    symbol.createSwatch(completion: { [weak self] (image, _) -> Void in
+                        // Make sure this is the cell we still care about and that it
+                        // wasn't already recycled by the time we get the swatch.
+                        if cell.tag != symbol.hashValue {
+                            return
+                        }
+                        
+                        // Set the swatch into our dictionary and reload the row
+                        self?.symbolSwatches[symbol] = image
```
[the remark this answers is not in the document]
mhdostal: Good point.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+            if let legendInfo = rowItem.content as? AGSLegendInfo,
+                let symbol = legendInfo.symbol {
+                if let swatch = self.symbolSwatches[symbol] {
+                    // We have a swatch, so set it into the imageView.
+                    legendInfoCell.symbolImage = swatch
+                } else {
+                    // Tag the cell so we know what symbol is being used
+                    // to create the swatch.  This will ensure we
+                    // put the legend image in the correct cell.
+                    cell.tag = symbol.hashValue
```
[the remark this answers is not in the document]
mhdostal: Ok, got it.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: I'm Ok with that.  @esreli ?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: I'm Ok with the renaming. @esreli ?

The reason they are separate subclasses is from the Toolkit Design Council.  They thought the API for these two common scenarios (Legend and TOC) would be more easily understood if there were classes for both.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
esreli: I like it!

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
esreli: @philium the subclasses were requested by Toolkit stakeholders. They were under the impression that our customers are accustomed to a **Legend** and a **TableOfContents** but not a **LayerContents**. While I agree with you here, because it's not hard to formalize these tools as subclasses, I say it's value added to make these types concrete.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,480 @@
...
+    override public init() {
+        super.init()
+        config = LayerContentsViewController.TableOfContents()
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
```
esreli, line 89: Can we consider supporting Storyboards?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,480 @@
...
+    private var symbolSwatches = [AGSSymbol: UIImage]()
+    
+    // The array of all layer contents to display in the table view.
+    private var displayedLayers = [AGSLayerContent]()
+    
+    // The array of all contents (`AGSLayer`, `AGSLayerContent`, `AGSLegendInfo`) to display in the table view.
+    private var contents = [Content]()
+    
+    // Array of child:parent used when determining accordion status.
+    private var parents = [UInt : [AGSLayerContent]]()
```
esreli, line 222: I haven't looked into this too deeply yet, though I always am skeptical of a child maintaining a reference to its parent. Is this potentially introducing a retain cycle?

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,126 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import ArcGIS
+import ArcGISToolkit
+
+private let layersExtra = "Layers"
```
esreli, line 19: These global variables can probably be computed from within their `enum Extras` counterpart. I would create another extension on `Extras` that supplies these titles.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: @philium Unless you have strong objections, because of the desire to have actual classes for the Legend and TableOfContents, I'd prefer to leave this the way it is.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/DataSource.swift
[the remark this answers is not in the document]
mhdostal: @philium @esreli 
I prefer it as it's own class, but I'm not averse to changing the name.  Does `LayerContentsDataSource` make sense?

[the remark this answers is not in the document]
esreli: Yes it does.

On the side... after learning more about the relationship between `LayerContentsTableViewController` and `LayerContentsViewController`, I would correct my above suggestion to `LayerContentsViewController`. Of course, this suggestion is moot as its own class.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/DataSource.swift
```diff
@@ -0,0 +1,122 @@
...
+    
+    /// Returns a `DataSource` initialized with the operational and base map layers of a
+    /// map or scene in an `AGSGeoView`.
+    /// - Parameter geoView: The `AGSGeoView` containing the map/scene's
+    /// operational and base map layers.
+    /// - Since: 100.8.0
+    public init(geoView: AGSGeoView) {
+        super.init()
+        self.geoView = geoView
+        geoViewDidChange()
```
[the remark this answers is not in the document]
mhdostal: I checked and it does *not* trigger the `didSet` when setting the `geoView` in `init`.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,126 @@
...
+    func userRequestsExtras(_ barButtonItem: UIBarButtonItem?) {
+        guard mapViewMode != .disabled else {
+            return
+        }
+
+        let action = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
+                
+        // Present the list of extras.
+        for extra in Extras.allCases {
+            var extraAction: UIAlertAction
```
philium, line 51: This doesn't need to be mutable.

```suggestion
            let extraAction: UIAlertAction
```

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,126 @@
...
+            action.addAction(extraAction)
+        }
+        
+        action.addAction(.cancel())
+        action.popoverPresentationController?.barButtonItem = barButtonItem
+        present(action, animated: true)
+    }
+    
+    func showLayerContents(_ barButtonItem: UIBarButtonItem?) {
+        var layerContentsVC: LayerContentsViewController
```
philium, line 72: This doesn't need to be mutable.

```suggestion
        let layerContentsVC: LayerContentsViewController
```

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,126 @@
...
+        }
+        
+        // Display the layerContentsVC as a popover controller.
+        layerContentsVC.modalPresentationStyle = .popover
+        layerContentsVC.popoverPresentationController?.barButtonItem = barButtonItem
+        present(layerContentsVC, animated: true)
+    }
+    
+    func showBookmarks(_ barButtonItem: UIBarButtonItem?) {
+        var bookmarksVC: BookmarksViewController
```
philium, line 93: This doesn't need to be mutable.

```suggestion
        let bookmarksVC: BookmarksViewController
```

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+        didSet {
+            legendImageView.image = symbolImage
+            symbolImage != nil ? activityIndicatorView.stopAnimating() : activityIndicatorView.startAnimating()
+        }
+    }
+    
+    var layerIndentationLevel: Int = 0 {
+        didSet {
+            // Set constraint constant; use +1 to account for initial indentation.
+            indentationConstraint.constant = CGFloat(layerIndentationLevel + 1) * indentationConstant
```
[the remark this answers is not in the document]
mhdostal: There's no limit on the sublayer count except what's supported by layers/web map.  I'm Ok without a limit and leaving it to the app/map developer, but I'll make the change it you want.

@philium, thoughts?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+let indentationConstant: CGFloat = 10.0
```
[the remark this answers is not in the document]
mhdostal: I move the cell reuse identifiers, but `indentationConstant` is used by both `UITableViewCells`.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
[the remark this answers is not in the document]
philium: @mhdostal I think I like the idea of leaving it up to the app developer to handle. Any limit we impose would just be arbitrary and could prevent legitimate use cases, like iPad-only apps that take advantage of the whole screen.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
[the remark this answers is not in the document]
philium: I don't see a problem with global variables as long as they're `private`. Could you please privatize this?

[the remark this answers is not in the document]
philium:
```suggestion
private let indentationConstant: CGFloat = 10.0
```

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: The relationship is only one way; child contain references to their parents, but the parents do not contain references to their children, so there should be no retain cycle.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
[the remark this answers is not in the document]
esreli: Sorry, I should have been more clear. The suggestion was not to limit the number of sublayers supported by the tool but rather limiting sublayer indentation at a certain depth.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,480 @@
...
+    internal struct Legend: LayerContentsConfiguration {
+        var layersStyle: ConfigurationStyle = .visibleLayersAtScale
+        var allowToggleVisibility: Bool = false
+        var allowLayersAccordion: Bool = false
+        var showSymbology: Bool = true
+        var respectInitialLayerOrder: Bool = false
+        var respectShowInLegend: Bool = true
+        var showRowSeparator: Bool = false
+        var title: String = "Legend"
+    }
```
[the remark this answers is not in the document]
mhdostal: The superclass (LayerContentsViewController) uses the `TableOfContents` struct as it's default; otherwise, I would agree.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/DataSource.swift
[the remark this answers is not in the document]
mhdostal: Renamed to `LayerContentsDataSource`

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,126 @@
...
+                extraAction = UIAlertAction(title: extra.description, style: .default, handler: { [weak self] (action) in
+                    self?.showLayerContents(barButtonItem)
+                })
+            case .bookmarks:
+                extraAction = UIAlertAction(title: extra.description, style: .default, handler: { [weak self] (action) in
+                    self?.showBookmarks(barButtonItem)
+                })
+            }
+            
+            action.addAction(extraAction)
```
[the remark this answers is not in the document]
mhdostal: I'm inclined to not include images, since they are not explicitly supported by Apple.

mhdostal:
@esreli, @philium I've addressed all the comments.  Unless I missed something, the only outstanding item item is:

- potentially show a title and "Done" button when displaying either the TOC or Bookmarks VC.

mhdostal:
@philium - can you give this a final(?) look?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+        // The layer is expanded.
+        case expanded
+        // The layer is collapsed.
+        case collapsed
+        // No accordion control.
+        case none
+    }
+
+    var kind: Kind = .layer
+    var content: AnyObject?
```
rolson, line 119: Content.content is a bit odd, is there a better name for the class or this variable?

rolson, line 119: instead of AnyObject, have you considered making Content generic?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+    var name: String = ""
+    var indentationLevel: Int = 0
+    var accordion: AccordionDisplay = .expanded
+    var allowToggleVisibility: Bool = false
+    var isVisibilityToggleOn: Bool = false
+    var isVisibleAtScale: Bool = true
+    var parents = [Content]()
```
rolson, line 126: does this create retain cycles?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+    }
+    
+    private var dataSourceObservation: NSKeyValueObservation?
+    
+    /// The `LayerContentsDataSource` specifying the list of `AGSLayerContent` to display.
+    /// - Since: 100.8.0
+    public var dataSource: LayerContentsDataSource? = nil {
+        didSet {
+            // Add an observer to handle changes to the dataSource.layerContents.
+            dataSourceObservation = dataSource?.observe(\.layerContents) { [weak self] (_, _) in
```
rolson, line 191: if this is just one to one communication, then a delegate method is clearer than KVO, IMO

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+                }
+                
+                contents += contentArray
+            }
+        }
+    }
+    
+    // MARK: - Utility
+    
+    /// Returns a unique UINT for each object. Used because AGSLayerContent is not hashable
```
rolson, line 459: UInt

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+    /// and we need to use it as the key in our dictionary of legendInfo arrays.
+    private static func objectIdentifierFor(_ obj: AnyObject) -> UInt {
```
rolson, line 461: I think you could instead use Hashable where you have AnyObject and use the hash for holding values in dictionaries.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+            !viewpoint.targetScale.isNaN {
+            return layerContent.isVisible(atScale: viewpoint.targetScale)
+        }
+        
+        return true
+    }
+}
+
+private extension AGSLayerContent {
+    func objectId() -> UInt {
```
rolson, line 477: If you don't go with the hashable approach, then I would call this `objectIdentifier` so it's not conflated with Feature objectIDs

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -0,0 +1,149 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import Foundation
+import ArcGIS
+import ArcGISToolkit
+
+/// Defines the Extras.
+public enum Extras : CaseIterable {
```
rolson, line 20: Is there a more descriptive name that we can use than Extras?

data-collection/data-collection/Assets.xcassets/caret-down.imageset/Contents.json
```diff
@@ -0,0 +1,26 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "caret-down@1x.png",
```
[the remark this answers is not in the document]
mhdostal: Those are from the Calcite Fonts [repo](https://github.com/ArcGIS/calcite-fonts) on GitHub.  Nathan created a script to generate iOS Assets from them.  I'll talk to @esreli ...

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsDataSource.swift
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// The list of all layers used to generate the TOC/Legend, read-only.  It contains both the
+    /// operational layers of the map/scene and the reference and base layers of the basemap.
+    /// The order of the layer contents is the order in which they are drawn
+    /// in a map or scene:  bottom up (the first layer in the array is at the bottom and drawn first; the last
+    /// layer is at the top and drawn last).
+    /// - Since: 100.8.0
+    @objc
+    public dynamic private(set) var layerContents = [AGSLayerContent]()
```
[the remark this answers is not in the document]
mhdostal: Yes, KVO support. That way the TOC knows when the list of layers gets updated.

data-collection/data-collection/Assets.xcassets/caret-down.imageset/Contents.json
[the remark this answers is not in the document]
esreli: My initial thought was the same as @rolson 's, the graphic looks a bit dated. Unfortunately, we can't use SF Symbols yet because it requires a min target of 13.0 and Data Collection supports the same min target as the Runtime.

Also, I really like the notion of promoting Esri assets as a foundation to OSA design . I did find another calcite graphic that might work for us, the "chevron". What do you think?

chevron-right@3x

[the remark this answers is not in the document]
rolson: That could work better. Can we see both in context and make a visual comparison?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsDataSource.swift
```diff
@@ -0,0 +1,123 @@
...
+    
+    private var mapOrSceneObservation: NSKeyValueObservation?
+
+    private func geoViewDidChange() {
+        mapOrSceneObservation?.invalidate()
+        if let mapView = geoView as? AGSMapView {
+            mapView.map?.load { [weak self] (error) in
+                guard let self = self else { return }
+                if let error = error {
+                    print("Error loading map: \(error)")
```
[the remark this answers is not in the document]
mhdostal: Not strictly necessary; you are correct, that the app should handle that type of error.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: No, because the parent doesn't have a reference to the children.

data-collection/data-collection/Assets.xcassets/caret-down.imageset/Contents.json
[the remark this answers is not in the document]
mhdostal: Caret - 
Caret

Chevron - 
Chevron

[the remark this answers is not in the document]
esreli: I like that it feels less heavy.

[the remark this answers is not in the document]
rolson: I like the Chevron much better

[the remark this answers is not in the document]
mhdostal: Ok, chevron it is.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+/// Defines how to display layers in the table.
+/// - Since: 100.8.0
+internal class Content {
```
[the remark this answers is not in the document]
mhdostal: How about `DisplayedContent`?  Any other suggestions?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: `object`?  So `Content.object` or `DisplayedContent.object`?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,486 @@
...
+    // Only displays layers that are in scale and visible.
+    case visibleLayersAtScale
+}
+
+/// Configuration is an protocol (interface) that drives how to format the layer contents table.
+/// - Since: 100.8.0
+public protocol LayerContentsConfiguration {
+    /// Specifies the `ConfigurationStyle` applied to the table.
+    /// - Since: 100.8.0
+    var layersStyle: ConfigurationStyle { get }
```
rolson, line 33: ConfigurationStyle should probably be nested under `LayerContentsConfiguration`. Without that, it's ambiguous

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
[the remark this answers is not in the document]
rolson: I'm looking at the doc for this and trying to come up with names. DisplayedContent doesn't seem better because if I'm looking at it from an app perspective, that is ambiguous. If it was nested under LayerContentsViewController that's better. But maybe only necessary if a better name can't be thought of.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
rolson: object, value - those might be better.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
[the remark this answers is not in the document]
rolson: Is this a "Row"? I didn't go very deep in the review so I'm not quite sure how this is used.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
philium: I like `object` better. Another idea is to combine this property with `kind` using an enum with associated values:

```swift
enum Object {
    case layer(AGSLayer)
    case sublayer(AGSLayerContent)
    case legendInfo(AGSLegendInfo)
}

File: data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift (hunk 42 above)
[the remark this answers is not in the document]
Comment (mhdostal) [reply data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift:95]: Yes, it is a Row in the Table.  `RowContent`?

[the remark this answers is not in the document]
Comment (rolson) [reply data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift:95]: The problem I'm looking to solve is that the fully qualified name of this type - if I look at it with all the other types in the app, do I know what it does? RowContent is better, but still lacks the "LayerContentsViewController" context IMO.

File: data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift [hunk 44]
```diff
@@ -0,0 +1,227 @@
...
+            if let legendInfo = rowItem.content as? AGSLegendInfo,
+                let symbol = legendInfo.symbol {
+                if let swatch = symbolSwatchesCache.object(forKey: symbol) {
+                    // We have a swatch, so set it into the imageView.
+                    legendInfoCell.symbolImage = swatch
+                } else {
+                    legendInfoCell.symbolImage = nil
+                    
+                    // We don't have a swatch for the given symbol, so create the swatch.
+                    symbol.createSwatch(completion: { [weak self] (swatch, _) -> Void in
```
rolson, line 173: I would have thought that the cell itself would call createSwatch and handle that internally, instead of a parent object doing it and setting it on the cell.

The way I've done this in the past is within the cell have a property for the cells' current "Value" or "Content", and on didSet call createSwatch. Within the completion of createSwatch check that your current value is the same as the value that you created the swatch for. Then there is no index matching, looking up indexPath for, etc. 

Also it compartmentalizes the fact that the cell does it's own operations.

However, I'm not dogmatic about this changing, but wanted to offer it as a suggestion.

You would probably need to pass a pointer to the cache when you create the cell so that the cell has access to the swatch cache.

philium: @rolson See previous discussion [here](https://github.com/Esri/data-collection-ios/pull/230#discussion_r412732348), [here](https://github.com/Esri/data-collection-ios/pull/230#discussion_r412746326), [here](https://github.com/Esri/data-collection-ios/pull/230), and [here](https://github.com/Esri/data-collection-ios/pull/230#discussion_r414875437).

rolson: ok - well it's fine the way it is then. Personally I think it's cleaner for the cell to hold the model and set it's UI. There are ways to solve the problem of it holding onto the model for too long I think.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: @philium - would you use the enum in place of both `kind` and `content`?  If so, how would that work?

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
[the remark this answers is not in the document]
mhdostal: Made the change to `LayerContentsRowConfiguration`.  It's explicit and states exactly what it is:  the configuration for a layer contents row in the table view.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: `LayerContentsConfiguration` is a protocol and the compiler complains when I try to put the enum there.  I changed the name to `LayerContentsConfigurationStyle`.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
mhdostal: @rolson Do either `MapViewExtras` or `MapViewControllerExtras` work?

cc: @esreli

rolson: Sure, the latter seems fine.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: I've updated the code to use the `enum Object` @philium suggested.  It cleaned up the `tableView:cellForRowAt:` method a bit.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: I tried `Hashable` briefly, but couldn't figure it out.  I did make the change to `objectId` -> `objectIdentifier` to be clearer.

mhdostal:
@philium, @rolson I made the changes requested, please have another look.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsDataSource.swift
```diff
@@ -56,10 +60,11 @@ public class LayerContentsDataSource: NSObject {
     /// in a map or scene:  bottom up (the first layer in the array is at the bottom and drawn first; the last
     /// layer is at the top and drawn last).
     /// - Since: 100.8.0
-    @objc
-    public dynamic private(set) var layerContents = [AGSLayerContent]()
+    public private(set) var layerContents = [AGSLayerContent]()
     
     private var mapOrSceneObservation: NSKeyValueObservation?
+    
+    public var delegate: LayerContentsDataSourceDelegate?
```
rolson, line 67: delegates should be weak usually to avoid retain cycle

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsDataSource.swift
```diff
@@ -102,11 +109,13 @@ public class LayerContentsDataSource: NSObject {
             // Append any reference layers to the `layerContents` array.
             if let referenceLayers = basemap.referenceLayers as? [AGSLayerContent] {
                 self.layerContents.append(contentsOf: referenceLayers)
+                self.delegate?.layerContentsDidChange(self.layerContents)
             }
             
             // Insert any base layers at the beginning of the `layerContents` array.
             if let baseLayers = basemap.baseLayers as? [AGSLayerContent] {
                 self.layerContents.insert(contentsOf: baseLayers, at: 0)
+                self.delegate?.layerContentsDidChange(self.layerContents)
```
rolson, line 118: You don't need this line everywhere you change layerContents. All you need to do is put this line in `didSet` on the layerContents property. Even if you do an `insert(...)`, `didSet` will get called IIRC.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -69,7 +69,7 @@ public protocol LayerContentsConfiguration {
 public class LegendViewController: LayerContentsViewController {
     override public init() {
         super.init()
-        config = LayerContentsViewController.Legend()
+        configuration = LayerContentsViewController.Legend()
```
rolson, line 72: probably can just say `.Legend()`

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -82,26 +82,26 @@ public class LegendViewController: LayerContentsViewController {
 public class TableOfContentsViewController: LayerContentsViewController {
     override public init() {
         super.init()
-        config = LayerContentsViewController.TableOfContents()
+        configuration = LayerContentsViewController.TableOfContents()
```
rolson, line 85: can probably just say `.TableOfContents()`

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
philium: Here's how you could conform `Kind` to `Hashable`:

```swift
extension LayerContentsRowConfiguration.Kind: Hashable {
    private var wrappedObject: AnyObject {
        switch self {
        case .layer(let layer):
            return layer
        case .sublayer(let layerContents):
            return layerContents
        case .legendInfo(let legendInfo):
            return legendInfo
        }
    }
    
    static func == (lhs: LayerContentsRowConfiguration.Kind, rhs: LayerContentsRowConfiguration.Kind) -> Bool {
        lhs.wrappedObject === rhs.wrappedObject
    }
    
    func hash(into hasher: inout Hasher) {
        hasher.combine(ObjectIdentifier(wrappedObject))
    }
}
```

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: I had tried that but the compiler won't let me, since `Legend` is not a property of `LayerContentsConfiguration` (which is the type for `configuration`).

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
mhdostal: same as above.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsDataSource.swift
mhdostal: I've incorporated both suggestions.  I needed to pass a completion handler into the `appendBasemap` function because it needs to load the baseman and so is async.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+        case .layer:
+            // rowItem is a layer or a sublayer which implements AGSLayerContent
+            let layerCell = (tableView.dequeueReusableCell(withIdentifier: LayerContentsTableViewController.layerCellReuseIdentifier) as! LayerCell)
+            cell = layerCell
+            setupLayerCell(layerCell, indexPath, rowItem)
+        case .sublayer:
+            // rowItem is a layer or a sublayer which implements AGSLayerContent
+            let layerCell = (tableView.dequeueReusableCell(withIdentifier: LayerContentsTableViewController.sublayerCellReuseIdentifier) as! LayerCell)
+            cell = layerCell
+            setupLayerCell(layerCell, indexPath, rowItem)
```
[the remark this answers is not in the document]
mhdostal: There are two cells, one for layers and one for sublayers, so that the client can customize the look for each in the storyboards.  The prototype cells currently differ in how the "name" label is formatted.

See the discussion [here](https://github.com/Esri/data-collection-ios/pull/230#discussion_r413069403).

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,528 @@
...
+        rowConfigurations.forEach { configuration in
+            // Get the array of parents for `content`.
+            let parentArray = parents[configuration.kind] ?? []
+            configuration.indentationLevel = parentArray.count
+            
+            // For each parent object, find the matching Content.
+            parentArray.forEach { parent in
+                rowConfigurations.forEach { potentialParentContent in
+                    // Search all contents to see if it's a match.
+                    if potentialParentContent.kind.matches(object: parent) {
```
philium, line 413: Since `Kind` conforms to `Equatable`, this check could just test for equality:

```suggestion
                    if potentialParentContent.kind == .sublayer(parent) {
```

And then `matches(object:)` wouldn't be needed.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,528 @@
...
+            // Filter any sublayers which are not visible or not showInLegend.
+            let sublayerContents = layerContent.subLayerContents.filter { ($0.isVisible &&
+                (configuration.respectShowInLegend ? $0.showInLegend : true)) || configuration.layersStyle == .allLayers}
+            
+            sublayerContents.forEach { subLayerContent in
+                let showAtScale = parentShowAtScale && shouldShowAtScale(subLayerContent)
+                
+                if (configuration.layersStyle == .visibleLayersAtScale && showAtScale) || configuration.layersStyle == .allLayers {
+                    let internalLegendInfos = legendInfos[layerContent.objectIdentifier()] ?? []
+                    let rowConfiguration = LayerContentsRowConfiguration(LayerContentsRowConfiguration.Kind.sublayer(subLayerContent),
```
philium, line 435:
```swift
                    let rowConfiguration = LayerContentsRowConfiguration(.sublayer(subLayerContent),
```

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,528 @@
...
+                    rowConfiguration.isVisibleAtScale = showAtScale
+                    rowConfigurations.append(rowConfiguration)
+                    updateLayerLegend(subLayerContent, parentShowAtScale: parentShowAtScale)
+                }
+            }
+        } else {
+            if let internalLegendInfos = legendInfos[layerContent.objectIdentifier()] {
+                let showAtScale = parentShowAtScale && shouldShowAtScale(layerContent)
+                let contentArray = internalLegendInfos.map { legendInfo -> LayerContentsRowConfiguration in
+                    let rowConfiguration = LayerContentsRowConfiguration(LayerContentsRowConfiguration.Kind.legendInfo(legendInfo),
```
philium, line 447:
```swift
                    let rowConfiguration = LayerContentsRowConfiguration(.legendInfo(legendInfo),
```

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,528 @@
...
+            title = configuration.title
+            generateLayerList()
+        }
+    }
+    
+    // The table view controller which displays the list of layers.
+    private var layerContentsTableViewController: LayerContentsTableViewController?
+    
+    // Dictionary of legend infos; keys are AGSLayerContent objectIdentifier values.
+    private var legendInfos = [UInt: [AGSLegendInfo]]()
```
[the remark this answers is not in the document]
mhdostal: That's an excellent idea, thanks!  Simplifies it a bunch.

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsViewController.swift
```diff
@@ -0,0 +1,517 @@
...
+        case let layerContent as AGSLayerContent:
+            kind = .sublayer(layerContent)
+        case let legendInfo as AGSLegendInfo:
+            kind = .legendInfo(legendInfo)
+        default:
+            kind = nil
+        }
+
+        return self == kind
+    }
```
philium, line 516: This method is no longer being used. Could you please remove it?

```suggestion
```

mhdostal:
@philium - Awesome, thank you for all the help!

philium:
@mhdostal You're welcome! I'm happy to help.

mhdostal:
@rolson Thanks!

PR #234, opened 2020-05-27
SPM Alamofire NetworkReachabilityManager

esreli:
A small PR that removes NetworkReachabilityManager source from project directory, introduces Alamofire as a SPM dependency and updates `AppReachability` to use the latest `NetworkReachabilityManager` API.


esreli:
https://github.com/Esri/data-collection-ios/pull/234#pullrequestreview-419682824

Confirmed, it does increase the app archive size:

`v.next` -> 781.5 MB
`eli/alamofire-spm` -> 790.3 MB

data-collection/data-collection/AppReachability.swift
```diff
@@ -26,20 +26,9 @@ extension NetworkReachabilityManager {
     ///
     /// - SeeAlso: AppContextChangeHandler.swift
     static let shared: NetworkReachabilityManager = {
-        
         guard let manager = NetworkReachabilityManager(host: AppConfiguration.basePortalDomain) else {
             fatalError("Network Reachability Manager must be constructed a valid service url.")
```
mhdostal, line 30:
```suggestion
            fatalError("Network Reachability Manager must be constructed using a valid service url.")
```

esreli:
I've run a quick analysis to determine the impact on export app sizes with thinning, averaged. The results indicate the app sizes are increased:

_(values in MB)_

| | v.next | eli/alamofire-spm | delta |
| --- | --- | --- | --- |
| compressed | 42.6 | 43.0 | +0.4 |
| uncompressed | 102.1 | 103.2 | +1.1 |

esreli:
> I like the fact that integration is very easy with SPM, but if we're just using one class then I agree with @philium that it does seem like a lot of overhead.

I agree, it does feel like a lot of overhead since most of the framework isn't used.

> Is there chance we'll use more of it in the future?

Because Runtime handles network requests it's unlikely, though I suppose it's not impossible. A viable candidate for using Alamofire in the future would under any circumstance that would normally leverage `URLSession`.

> Is the example of how to incorporate a Swift Package part of the appeal?

I would say there is a little appeal in leveraging Swift Packages in Xcode 11. More than that, however, this Alamofire network utility can now treated like a first class dependency.

---

I suppose the question we're trying to answer here is choosing between a substantial improvement to dependency management vs. a moderate size increase of app archives and a negligible size increase of app exports.

PR #236, opened 2020-06-03
Swaps 'AppContainer' & 'Drawer' for 'Profile'

esreli:
This PR updates the app architecture to reflect the UX more familiar to iOS users and in-line with the Apple HIG.

In order to do so a number of tasks were required.

- Remove AppContainer paradigm
- Remove Drawer paradigm
- Introduce Profile paradigm
- Specify Map as main storyboard

Along the way a number of other tasks were accomplished

- Swaps some icons for calcite icons
- Slight project reorganization.

Note, this PR seems vast. In some ways it is though in other ways there are a lot of small file changes. After realizing this I think the PR will be less daunting.

| Old Map View | AppContainer & Drawer |
| -------------- | ------------------------|
| Screen Shot 2020-06-03 at 2 18 24 PM |  Screen Shot 2020-06-03 at 2 18 31 PM |

| New Map View | Profile |
| ---------------| -------|
| Screen Shot 2020-06-03 at 2 16 35 PM | Screen Shot 2020-06-03 at 2 16 43 PM |

Additionally, the Profile view supports Dark Mode and Accessible Dynamic Text resizing by default.

| Dark Mode | Dynamic Text |
| --- | --- |
| Screen Shot 2020-06-03 at 2 22 47 PM | Screen Shot 2020-06-03 at 4 15 26 PM |


data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
```diff
@@ -39,7 +39,7 @@ class MapViewController: UIViewController {
         case offlineMask
     }
 
-    weak var delegate: MapViewControllerDelegate?
+//    weak var delegate: MapViewControllerDelegate?
```
rolson, line 42: old code?

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+LocationSelection.swift
```diff
@@ -198,13 +198,26 @@ extension MapViewController {
...
-            present(simpleAlertMessage: "Could not determine extent for offline map.")
+            present(simpleAlertMessage: error.localizedDescription)
+            return
         }
+        
+        let scale = map.minScale
+        let directory: URL = .temporaryOfflineMapDirectoryURL(forWebMapItemID: .webMapItemID)
+        let offlineJob = OfflineMapJobConstruct.downloadMapOffline(map, directory, geometry, scale)
+        
+        EphemeralCache.set(object: offlineJob, forKey: OfflineMapJobConstruct.EphemeralCacheKeys.offlineMapJob)
```
rolson, line 219: `set` is an "overly vacuous" name according to swift docs. The function should be `setObject(_:forkey)`.

Also, you might want to make it an instance method on a shared cache:
`EphemeralCache.shared.setObject(...)`

rolson: Also it would be nice if you could just do this:
`.offlineMapJob`
instead of 
`OfflineMapJobConstruct.EphemeralCacheKeys.offlineMapJob`

to do that, extend the key type with a static var called `offlineMapJob`

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Map.swift
```diff
@@ -19,21 +19,17 @@ extension MapViewController {
...
-            
-            guard let self = self else { return }
-            
-            guard error == nil else {
-                let error = (error! as NSError)
-
+        func handleMapLoadCallback(error: Error?) -> Void {
+            // If there's an error loading the map we want to inform the user and disable the map.
+            if let error = error {
+                let error = error as NSError
```
rolson, line 32: i think you can combine the above 2 lines:
`if let error: NSError = error`

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Map.swift
```diff
@@ -74,25 +70,23 @@ extension MapViewController {
             }
             
             self.mapViewMode = .defaultView
+            self.title = map.item?.title ?? "Map"
+            let layers: [AGSFeatureLayer] = map.operationalLayers.compactMap { $0 as? AGSFeatureLayer }
```
rolson, line 74: the type annotation isnt necessary:
```
let layers = map.operationalLayers.compactMap { $0 as? AGSFeatureLayer }
```

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Map.swift
```diff
@@ -74,25 +70,23 @@ extension MapViewController {
...
             
-            self.delegate?.mapViewController(self, didUpdateTitle: map.item?.title ?? "Map")
-            
-            guard let operationalLayers = map.operationalLayers as? [AGSFeatureLayer] else {
-                self.delegate?.mapViewController(self, shouldAllowNewFeature: false)
-                return
-            }
-            
-            AGSLoadObjects(operationalLayers, { [weak self] (_) in
+            AGSLoadObjects(layers) { [weak self] (_) in
```
rolson, line 76:
```
AGSLoadObjects(layers) { [weak self] (_) in
```

could probably just be

```
AGSLoadObjects(layers) { [weak self] _ in
```

data-collection/data-collection/Utilities/EphemeralCache.swift
```diff
@@ -20,7 +20,7 @@ import Foundation
 /// - Note: The `EphemeralCache` uses `NSCache` and a concurrent `DispatchQueue` (with a barrier flag).
 class EphemeralCache {
```
rolson, line 21: The naming of the functions in the class don't really follow [swift naming guidelines](https://swift.org/documentation/api-design-guidelines/). I would copy naming guidelines in the `Dictionary` class.

an example is `set` is "absurdly vacuous": 

> Never transform a name into "get", "set", "with", "for", or "using", just to avoid creating absurdly vacuous names.

from [here](https://github.com/apple/swift-evolution/blob/master/proposals/0005-objective-c-name-translation.md)

data-collection/data-collection/Utilities/EphemeralCache.swift
```diff
@@ -45,21 +45,21 @@ class EphemeralCache {
...
     
-    static func remove(objectForKey key: String) {
-        shared[key] = nil
+    func remove(objectForKey key: String) {
+        self[key] = nil
     }
     
-    static func has(objectForKey key: String) -> Bool {
-        return shared[key] != nil
+    func has(objectForKey key: String) -> Bool {
```
rolson, line 62: should probably be "contains"

PR #237, opened 2020-06-04
Fixes bug, renames, and updates `AddressLocator`.

esreli:
I discovered a bug in the implementation of `AGSLoadableBase`.

When a user loaded the `AddressLocator` (ne `AppReverseGeocoderManager`), the tool would load both online and offline locators in one fell swoop. This is flawed because if the user was online, reachable and wanted to perform a reverse geocoder operation on the world geocoder service, if the offline locator did not load properly, the tool would entirely fail with a `loadStatus` of `.notLoaded`.

This is less desire-able because the tool should allow either locator to function even if the counterpart fails.

I used the PR to improve the tool as well.

- New name is more succinct and describes better its function.
- Lazily loads locators.
- Creates `String` extensions rather than `struct Keys` for attribute keys and offline locator name.
- Removes unneeded/unnecessary `AGSLoadableBase` subclass implementation.


data-collection/data-collection/Address Locator/AddressLocator.swift
```diff
@@ -0,0 +1,89 @@
...
+                return params
+            }()
+            // Perform the reverse geocode task.
+            locator.reverseGeocode(withLocation: point, parameters: params) { (results, error) in
+                if let error = error {
+                    completion(.failure(error))
+                }
+                else if
+                    let attributes = results?.first?.attributes,
+                    let address = attributes[.address] as? String ?? attributes[.matchAddress] as? String {
```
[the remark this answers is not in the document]
esreli: It does work, nice recommendation.

data-collection/data-collection/Address Locator/AddressLocator.swift
```diff
@@ -0,0 +1,89 @@
...
+        if appContext.workMode == .online && appReachability.isReachable {
+            locator = onlineLocator
+        }
+        // Otherwise, we'll use the offline locator.
+        else {
+            locator = offlineLocator
+        }
+        // Load the chosen locator.
+        locator.load { (error) in
+            if let error = error {
```
philium, line 46: I commented this (or at least tried to) on my first review, but it got lost. Should the work mode and reachability be rechecked here? I'm just wondering what would happen if the locator wasn't loaded and while it is loading, things change such that the other locator should be used by the time the completion handler is called.

esreli: This is a good point. I've written a simple implementation that fails `reverseGeocodeAddress` if the app context has switched since the load.

A more robust implementation exists out there but i'm not sure if this scenario merits the extra effort. Do you? Open to ideas here.

data-collection/data-collection/Address Locator/AddressLocator.swift
```diff
@@ -0,0 +1,97 @@
...
+    ///   - point: The point used in the reverse geocode operation.
+    ///   - completion: A closure called upon completion of the reverse geocode.
+    ///   - result: The result of the reverse geocode operation, with either the
+    ///   address or an error.
+    func reverseGeocodeAddress(for point: AGSPoint, completion: @escaping (_ result: Result<String, Error>) -> Void) {
+        let locator = appContextAwareLocator
+        locator.load { [weak self] (error) in
+            // Ensure the loaded locator matches the app context aware locator.
+            // The app context might have changed since the locator started loading.
+            guard locator == self?.appContextAwareLocator else {
```
philium, line 49: I don't think `AGSLocatorTask` actually implements `isEqual(_:)`, so you'll want to use pointer comparison (`===`).

data-collection/data-collection/Address Locator/AddressLocator.swift
philium: Whatever you think is best.

data-collection/data-collection/Address Locator/AddressLocator.swift
esreli: > I don't think `AGSLocatorTask` actually implements `isEqual(_:)`, so you'll want to use pointer comparison (`===`).

I believe this is the default behavior already for `NSObject`. According to the [Obj-C docs](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/Introspection/Introspection.html#//apple_ref/doc/uid/TP40010810-CH9-SW61):

> The default `NSObject` implementation of `isEqual:` simply checks for pointer equality.

Because `AGSLocatorTask` doesn't implement `isEqual(_:)`, (`==`) will fallback to pointer comparison, no? Am I missing something?

philium: Right! I knew that. I think.

PR #239, opened 2020-06-26
Unifies API for storing and retrieving `AGSJSONSerializing` objects to/from UserDefaults

esreli:
This PR unifies the API for storing and retrieving `AGSJSONSerializing` objects to/from UserDefaults using generics.

Previously a user would:

```swift
UserDefaults.standard.set(newValue, forKey: visibleAreaDefaultsKey)
AGSViewpoint.retrieveFromUserDefaults(forKey: visibleAreaDefaultsKey)
```

This feels awkward, as if you have to use two separate APIs for the storage and retrieval. Using generics and implicit types, we can modify this API such that both storing and retrieving methods are extensions on `UserDefaults`, unifying the API.

```swift
UserDefaults.standard.setSerializable(newValue, forKey: visibleAreaDefaultsKey)
UserDefaults.standard.getSerializable(forKey: visibleAreaDefaultsKey)
```


data-collection/data-collection/Extensions/Foundation/UserDefaults+AGSJSONSerializable.swift
```diff
@@ -40,5 +38,23 @@ extension UserDefaults {
...
+    ///
+    func getSerializable<Serializable: AGSJSONSerializable>(forKey key: String) -> Serializable? {
+        do {
+            guard let json = UserDefaults.standard.value(forKey: key) else { return nil }
+            return try Serializable.fromJSON(json) as? Serializable
+        }
+        catch {
+            print("[Error: AGSJSONSerializable] could not serialize object from JSON.", error.localizedDescription)
+        }
+        return nil
```
philium, line 58: My only suggestion would be to move this inside the `catch` clause since that's the only case where it can be reached.

PR #240, opened 2020-06-26
Converts `AppFiles` to `FileManager` extension

esreli:
Converts `AppFiles` type to `FileManager` extension, accomplishing the same task.

### DIFF
--- a/data-collection/data-collection/AppFiles.swift
+++ b/data-collection/data-collection/AppFiles.swift
@@ -15,7 +15,7 @@
 import Foundation
 import ArcGIS
 
-/// This class extension is used to creating and removing directories and items needed for the app in the device's file documents directory.
+// This class extension is used to creating and removing directories and items needed for the app in the device's file documents directory.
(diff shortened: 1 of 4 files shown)


data-collection/data-collection/AppFiles.swift
```diff
@@ -15,29 +15,26 @@
 import Foundation
 import ArcGIS
 
-/// This class is used for creating and removing directories and items needed for the app in the device's file documents directory.
-class AppFiles {
-    
-    private var fm: FileManager { return FileManager.default }
-    
+/// This class extension is used to creating and removing directories and items needed for the app in the device's file documents directory.
```
philium, line 18: As far as I know, there's no benefit to using a documentation comment on extensions.

```suggestion
// This class extension is used to creating and removing directories and items needed for the app in the device's file documents directory.
```

esreli: Good catch.

PR #241, opened 2020-06-29
`UserDefaults` supporting `RawRepresentable` storage & retrieval 

esreli:
This PR deprecates the `AGSPopupAttachmentSize` type-specific `UserDefaults` storage mechanism. Now the app opts to use a new generic mechanism for storing and retrieving `RawRepresentable` types.

## Known limitation.

This PR facilitates the storage and retrieval of `RawRepresentable` types but makes no inference for the `RawValue` of the `RawRepresentable`. It is the dev's responsibility to ensure the underlying `RawValue` is `UserDefaults` compliant.


data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupEditStagedPhotoAttachmentViewController/RichPopupEditStagedAttachmentViewController.swift
```diff
@@ -137,7 +137,10 @@ class RichPopupEditStagedAttachmentViewController: UITableViewController, UIText
                 stagedAttachment.preferredSize = AGSPopupAttachmentSize.tableViewMap[indexPath.row]
                 
                 // Store the selected value in `UserDefaults`.
-                stagedAttachment.preferredSize.storeDefaultPopupAttachmentSize()
+                UserDefaults.standard.set(
+                    stagedAttachment.preferredSize,
+                    forKey: .defaultAttachmentSize
```
philium, line 142: Rather than exposing the key, you could just add a property to `UserDefaults`:

```swift
extension UserDefaults {
    var defaultAttachmentSize: AGSPopupAttachmentSize? {
        get {
            UserDefaults.standard.rawRepresentable(forKey: .defaultAttachmentSize)
        }
        set {
            UserDefaults.standard.set(newValue, forKey: .defaultAttachmentSize)
        }
    }
}
```

And then keep the key a private implementation detail.

data-collection/data-collection/Extensions/Foundation/UserDefaults+RawRepresentable.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// - Parameters:
+    ///     - forKey: `UserDefaults` key.
+    ///
+    func rawRepresentable<RR: RawRepresentable>(forKey key: String) -> RR? {
+        if let value = value(forKey: key) as? RR.RawValue {
+            return RR(rawValue: value)
+        }
+        else {
+            return nil
+        }
```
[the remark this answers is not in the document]
esreli: Okay, this can be reduced to a single line though I think we should examine the definition of "simplified" ;) I would suggest the solution I've presented is considerably more readable and solves the problem using similar logic. Can we leave as-is?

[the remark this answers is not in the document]
philium: It is up to you. When I first learned about `Optional.map(_:)` and `Optional.flatMap(_:)`, I didn't see how they were useful or why they are in the standard library. But once I started using them, I've found more and more places where they can be helpful. As far as readability, I've actually found them to be more readable as the `else` case (actually `.none` [case](https://github.com/apple/swift/blob/d2b68c70dadb46ee22ef8dff36ab92e20fe0dc51/stdlib/public/core/Optional.swift#L198), but whatever) is always the same: `nil`. Using an `if` statement, the `else` clause must be read to understand what happens in the case that the optional binding fails.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/Extensions/AGSPopupAttachmentSize+PreferredSize.swift
```diff
@@ -14,17 +14,17 @@
...
-        UserDefaults.standard.set(self.rawValue, forKey: AGSPopupAttachmentSize.userDefaultsKey)
-    }
-    
-    static func retrieveDefaultPopupAttachmentSize() -> AGSPopupAttachmentSize {
-        
-        let storedRaw = UserDefaults.standard.integer(forKey: AGSPopupAttachmentSize.userDefaultsKey)
-        return AGSPopupAttachmentSize(rawValue: storedRaw) ?? .actual
+extension String {
+    fileprivate static let defaultAttachmentSize = "AGSPopupAttachmentSize.PreferredAttachmentSize.UserDefaultKey"
+}
```
philium, line 19: Since `private` and `file private` are the same at the top level of a file, it is recommended to just use `private` and use `fileprivate` only when `private` won't cut it. Also, in general, I recommend putting the access control keyword on the extension when possible. This makes it very handy to add additional private members in the future to extensions like this, where all the members are just private implementation details.

```suggestion
private extension String {
    static let defaultAttachmentSize = "AGSPopupAttachmentSize.PreferredAttachmentSize.UserDefaultKey"
}
```

data-collection/data-collection/Extensions/Foundation/UserDefaults+RawRepresentable.swift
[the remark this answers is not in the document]
esreli: I tend agree with you, `Optional.map(_:)` does improve readability here. I'm trying to reconcile that idea with the parentheses soup cooked up in the suggestion. To me it would feel more readable if it used shorthand argument names.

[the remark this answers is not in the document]
philium: Just keep in mind that wrapping the call to the initializer in a closure like that is less efficient. Passing the initializer directly prevents the creation of an intermediate closure.

data-collection/data-collection/Extensions/Foundation/UserDefaults+RawRepresentable.swift
```diff
@@ -0,0 +1,40 @@
...
+    
+    /// Write a `RawRepresentable` object to `UserDefaults`.
+    ///
+    /// Extract the object's `RawValue` and store it in `UserDefaults`.
+    ///
+    /// - Parameters:
+    ///     - forKey: `UserDefaults` key.
+    ///
+    func setRawRepresentable<RR: RawRepresentable>(_ value: RR?, forKey key: String) {
+        set(value?.rawValue, forKey: key)
```
philium, line 27: I just realized that nowhere is the raw value type constrained in any way, so I looked at the documentation for `NSUserDefaults.set(_:forKey:)` and found the following:

> The value parameter can be only property list objects: `NSData`, `NSString`, `NSNumber`, `NSDate`, `NSArray`, or `NSDictionary`. For `NSArray` and `NSDictionary` objects, their contents must be property list objects.

In other words, this method cannot be used with a value of just any type that conforms to `RawRepresentable`. As such, you might consider constraining `RawValue`. If the major use case is for integer-backed enums, then you could add a requirement that `RR.RawValue == Int`.

Same for the getter.

esreli: I don't think I did a good enough job alluding to this in the **known limitation** section of the PR's description.

I didn't feel the need to constrain the value to supported types under the assumption that the code must follow the same rules as it normally does when using the `UserDefaults` API, instead we swap setting/getting the object for setting/getting the object's `rawValue`. This extension essentially serves as a proxy between the `RawRepresentable` and `UserDefaults` with the assertion that `RawValue` must play by the same rules required by the `UserDefaults` API. 

I think this is a fair assumption.

philium: Oops. I totally missed the "Known limitation" section. That explanation would be very helpful to put in a documentation comment for both methods.

Issue #229, opened 2020-03-27
remove app secrets from versioned source code


jf990:
Regarding https://github.com/Esri/data-collection-ios/blob/master/data-collection/data-collection/AppConfiguration.swift#L66

`licenseKey` and `clientID` should not be managed in AppConfiguration.swift, as a developer would never want to include these values in source control. Suggesting an app configuration this way would not be considered a best practice.

We should also remove our clientID from source as a hacker could use it to deploy an app. Or consider using a rate limited proxy instead.

PR #244, opened 2020-07-01
Hotfix: App Secrets

esreli:
This PR is an attempt to resolve https://github.com/Esri/data-collection-ios/issues/229 once and for all.

This technique uses Xcode Environment Variables to store sensitive keys, supplying these keys at build time. The main **benefit** is that this technique removes sensitive keys from version control. The main **drawback** is that this technique requires an additional step before new users can run the application.


esreli:
I've given this technique more thought and it will not satisfy our needs. I had mistakenly believed the parameters were packaged with the app and sent at runtime, even on device. Something felt a bit off here and so I looked into it a bit deeper. I have confirmed that sending environment variables from an Xcode scheme is limited to the owner of the scheme. This technique will not transfer outside of a dev or CI environment and into production. Closing.

jf990:
@esreli well, it appeared promising 🤷‍♂️

esreli:
@jf990 @nixta  i've submit a second attempt to solve this problem that uses values stored in a file rather than Xcode Environment variables.

https://github.com/Esri/data-collection-ios/pull/246

PR #245, opened 2020-07-16
Add TOCs and rename detailed documentation

mikewilburn:
This PR uses the [atom-mdtoc](https://atom.io/packages/atom-mdtoc) package for Atom to introduce tables of contents at the top of our two documentation markdowns - the root README.md as well as the one located at docs/index.md.

It also ensures the detailed documentation is directly referenced from within the README at the repository's root.

Lastly, it renames docs/index.md to docs/README.md to leverage automatic rendering within that directory.


mikewilburn:
Fyi, looks like my commits inadvertently removed a number of trailing spaces within the affected documents, so be aware that it may appear as though more lines have changed than in actuality.

README.md
```diff
@@ -26,6 +47,9 @@ The project also demonstrates some patterns for building real-world apps around
 * Swift class extensions
 * Swift KVO & NotificationCenter state handling
 
+## Detailed Documentation
+See the docs [README.md](./docs/README.md) for a detailed explanation of the application, including its architecture and how it leverages the ArcGIS platform, as well as how you can begin using the app right away.
```
esreli, line 51: I think we should hyperlink to the "docs" rather than the docs "README".. it seems to make more sense to me for the same reason we're updating the name of the file within the docs directory, so we can simply refer to it as "the docs".

```suggestion
Read the [docs](./docs/README.md) for a detailed explanation of the application, including its architecture and how it leverages the ArcGIS platform, as well as how you can begin using the app right away.
```

README.md
```diff
@@ -1,5 +1,26 @@
...
+      - [1. Register an Application](#1-register-an-application)   
+      - [2. Configuring the project](#2-configuring-the-project)   
+- [Learn More](#learn-more)   
+- [Requirements](#requirements)   
+- [Contributing](#contributing)   
+- [Licensing](#licensing)   
+   - [3rd Party Component Licensing](#3rd-party-component-licensing)   
+
+<!-- /MDTOC -->
+
```
esreli, line 23: Can we put a horizontal rule here?
```suggestion
---
```

RELEASE.md
```diff
@@ -1,3 +1,8 @@
+# Release 1.2.1
+
+- Adds table of contents to root README.md and docs/index.md
```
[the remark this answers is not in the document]
mikewilburn: I'm hoping that it should be clear, but it certainly doesn't hurt to add it. I think it's a reasonable suggestion. 👍

mikewilburn:
@esreli I think I took care of all of your suggestions. Have another look and see what you think? Let me know if you think where I added mention of the MDTOC extension in the root README makes sense.

> ### Comment: MDTOC [...]
> 
> I take it this comment is used by the atom extension for generating a TOC with the same parameters from person to person? Is it such that the TOC is generated just under the comment?

Yes, this is exactly correct. Changing any of the in-line markdown comments for the MDTOC parameters will affect the way the table of contents generates from that point onward.

PR #247, opened 2020-07-31
Mhd/194 floating panel

mhdostal:
This is the FloatingPanel implementation for DataCollection v1.3.  It also moves the Bookmarks from their own popover to a floating panel (LayerContents and Identify results will move later).

In addition to a code-level review, I'd be interested to hear how the floating panel "feels" when resizing and switching between visible states (minimized, partial, full).


mhdostal:
cc: @philium @pgruenler @rolson @nCastle1 @puneet-pdx @njarecha @sbaskaran - if you have a chance, please review.  I need at least one more reviewer.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+        updateImage()
+        updateCloseButton()
+        updateHeaderStackView()
+        updateAllowManualResize()
+        updateFloatingPanelTitle()
+        updateFloatingPanelSubtitle()
+    }
+    
+    /// Sets up the constrains for both regular and compact horizontal size classes.
+    /// - Parameter superview: The superview of the floating panel.
```
philium, line 308:
```suggestion
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+    private func setupConstraints() {
```
philium, line 309: I don't know if you explored it as an option, but another way of sizing the floating panel would be to use intrinsic content size. It seems like the panel has an intrinsic height In compact width environments and both an intrinsic height and width in regular width environments. I think you can even invalidate intrinsic content size inside of an animation block to animate intrinsic content size changes. I can't say for sure, but that might have been a simpler option.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+        // updateState will update the resizableLayoutConstraint constant
+        // based on the current state.
+        updateState(false)
+        
+        // Define the constraints used for a regular-width layout.
+        regularWidthConstraints = [
+            view.leadingAnchor.constraint(equalTo: superview.safeAreaLayoutGuide.leadingAnchor, constant: regularWidthInsets.left),
+            view.widthAnchor.constraint(equalToConstant: floatingPanelWidth),
+            view.topAnchor.constraint(equalTo: superview.safeAreaLayoutGuide.topAnchor, constant: regularWidthInsets.top),
+            view.bottomAnchor.constraint(lessThanOrEqualTo: superview.safeAreaLayoutGuide.bottomAnchor, constant: -regularWidthInsets.bottom),
```
philium, line 325: I personally find negative constants a little harder to reason about, so I avoid them whenever possible by putting the trailing-most (or bottom-most) view on the left hand side of the constraint equation. For example, `view.bottomAnchor` and `superview.safeAreaLayoutGuide.bottomAnchor` could be swapped to make the constant positive:

```suggestion
            superview.safeAreaLayoutGuide.bottomAnchor.constraint(greaterThanOrEqualTo: view.bottomAnchor, constant: regularWidthInsets.bottom),
```

But that's just a preference. You don't have to do it that way.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+        resizeableLayoutConstraint = view.heightAnchor.constraint(equalToConstant: 0)
+        resizeableLayoutConstraint.priority = .defaultLow
+        
+        // updateState will update the resizableLayoutConstraint constant
+        // based on the current state.
+        updateState(false)
+        
+        // Define the constraints used for a regular-width layout.
+        regularWidthConstraints = [
+            view.leadingAnchor.constraint(equalTo: superview.safeAreaLayoutGuide.leadingAnchor, constant: regularWidthInsets.left),
```
philium, line 322: I am a bit surprised to see the view controller positioning its view within its superview. That's rather unusual. I can't help but wonder if there's a way for the super view [controller] to position the floating panel. Something like so:

```swift
extension UIViewController {
    func showInFloatingPanel(_ viewController: UIViewController) { ... }
    func hideFloatingPanel() { .. }
}
```

I haven't fully thought through that yet, but it seems like it would be possible. The nice thing about that API is that it enforces there being only one floating panel at a time. If one is already presented, calling showInFloatingPanel(_:)` would simply replace the view controller inside the panel with the new one.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+    }
+    
+    /// Handles the pan gesture used to resize the floating panel.
+    /// - Parameter gestureRecognizer: The active gesture recognizer.
+    @IBAction private func handlePanGesture(_ gestureRecognizer : UIPanGestureRecognizer) {
+        // Get the changes in the X and Y directions relative to
+        // the superview's coordinate space.
+        let translation = gestureRecognizer.translation(in: gestureRecognizer.view?.superview)
+        
+        if gestureRecognizer.state == .began {
```
philium, line 378: The recognizer's state is checked here, on line `386`, and again on line `404`. Each state is mutually exclusive. Could you please use a switch statement instead?

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+            // Save the view's original constant.
+            initialResizableLayoutConstraintConstant = resizeableLayoutConstraint.constant
+        }
+        
+        // Velocity will be used to determine whether to handle
+        // the 'flick' gesture to switch between states.
+        let velocity = gestureRecognizer.velocity(in: gestureRecognizer.view?.superview)
+        if gestureRecognizer.state == .ended {
+            if abs(velocity.y) > 0 {
```
philium, line 387: In most cases, instance methods are preferred over global functions.
```suggestion
            if velocity.y.magnitude > 0 {
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+        return partialHeight > 0 ? partialHeight : maximumHeight * 0.40
+    }
+    
+    /// The minimum height of the floating panel, taking into account the
+    /// horizontal size class.
+    private var minimumHeight: CGFloat {
+        let headerBottom = headerStackView.frame.origin.y + headerStackView.frame.size.height
+        let handlerbarHeight = bottomHandlebarView.frame.height
+        
+        // For compactWidth, handlebar is on top, so headerBottom is the limit;
```
philium, line 87:
```suggestion
        // For compactWidth, handlebar is on top, so headerBottom is the limit.
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+        // Skip if already initialized or we don't have a superview.
+        guard !initialized else { return }
+        setupConstraints()
+        initialized = true
+    }
+    
+    /// Update constraints for current horizontal size class.
+    private func updateInterfaceForCurrentTraits() {
+        NSLayoutConstraint.deactivate(regularWidthConstraints)
+        NSLayoutConstraint.deactivate(compactWidthConstraints);
```
philium, line 360:
```suggestion
        NSLayoutConstraint.deactivate(compactWidthConstraints)
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+        
+        NSLayoutConstraint.activate(isCompactWidth ? compactWidthConstraints : regularWidthConstraints);
```
philium, line 362:
```suggestion
        NSLayoutConstraint.activate(isCompactWidth ? compactWidthConstraints : regularWidthConstraints)
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+            // Save the view's original constant.
+            initialResizableLayoutConstraintConstant = resizeableLayoutConstraint.constant
+        }
+        
+        // Velocity will be used to determine whether to handle
+        // the 'flick' gesture to switch between states.
+        let velocity = gestureRecognizer.velocity(in: gestureRecognizer.view?.superview)
+        if gestureRecognizer.state == .ended {
+            if abs(velocity.y) > 0 {
+                handleFlick(velocity);
```
philium, line 388:
```suggestion
                handleFlick(velocity)
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+            }
+            else if resizeableLayoutConstraint.constant < minimumHeight {
+                // Constraint is less than minimum.
+                state = .minimized
+                resizeableLayoutConstraint.constant = minimumHeight
+            }
+            else if resizeableLayoutConstraint.constant > maximumHeight {
+                // Constraint is greater than maximum.
+                state = .full
+                resizeableLayoutConstraint.constant = maximumHeight
```
philium, line 398: What is the source of truth for the state of the panel? Here `resizeableLayoutConstraint.constant` is set after the state is set, implying that `state` is the source of truth. But in `handleFlick(_:)`, `state` is set based on the value of `resizeableLayoutConstraint.constant`.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+            }
+            else {
+                // The user is dragging the view shorter.
+                // Switch to either .minimized or .partial depending on the constraint constant value.
+                state = resizeableLayoutConstraint.constant <= internalPartialHeight ? .minimized : .partial
+            }
+        }
+    }
+    
+    public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
```
philium, line 458: image
```suggestion
    public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
        super.traitCollectionDidChange(previousTraitCollection)
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+        // Complete all pending layout operations.
+        superview.layoutIfNeeded()
+        
+        // Animate the transition to the new constraint constant value.
+        UIView.animate(withDuration: 0.2,
+                       delay: 0,
+                       usingSpringWithDamping: 0.85,
+                       initialSpringVelocity: 25,
+                       options: [.layoutSubviews, .curveEaseInOut],
+                       animations: { [weak self] in
```
philium, line 478: `weak self` isn't needed here since there isn't a retain cycle.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+    public var floatingPanelSubtitle: String? = "Subtitle" {
+        didSet {
+            guard isViewLoaded else { return }
+            updateFloatingPanelSubtitle()
+        }
+    }
+    
+    /// The initial view controller to display in the content area of
+    /// the floating panel view controller.
+    public var initialViewController: UIViewController? {
```
philium, line 219: This name seems a bit odd to me since there's only one view controller, right? So how is it the initial one?

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.storyboard
```diff
@@ -0,0 +1,257 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
philium, line 1: I'm a bit surprised to see the panel's view implemented in a storyboard rather than a nib since it is just one view. Storyboards are great for multiple views and modeling connections between them, but for a single view, a nib can be much better. For one thing, the view controller could be instantiated directly rather than through the `UIStoryboard` APIs. So you could use `FloatingPanelViewController(initialViewController: bookmarksViewController)` and have `initialViewController` be immutable. There are some things that can't be done with a nib (like dynamic table view cells), but it seems like a nib would work here. Just a thought.

nCastle1:
I've taken a look and have a few comments on the behavior of the bottom sheet:

* When showing just the title/preview for the bottom sheet, I think the content of the sheet should be hidden. Currently on iPhone XR, the bookmark list is visible behind the handlebar/in the unsafe area when in portrait. Apple maps will hide the sheet content even when it would otherwise be visible behind the handlebar.
* When the bottom sheet is expanded and the user pulls down on the bookmark list, the bottom sheet doesn't start transitioning to the intermediate state. Apple maps will start panning the sheet once the list can't scroll anymore. 
* Noticed the bottom sheet currently fades out. I wonder if it would be more natural if it slide in from the bottom when shown at the bottom, and from the left when shown on the side?

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+/// The protocol implemented to respond to floating panel view controller events.
+public protocol FloatingPanelViewControllerDelegate: AnyObject {
+    /// Tells the delegate that the user has requested to dismiss the floating panel view controller.
+    ///
+    /// - Parameter floatingPanelViewController: The floating panel view controller to dismiss.
+    func userDidRequestDismissFloatingPanel(_ floatingPanelViewController: FloatingPanelViewController)
+}
+
+/// Defines the vertical state of the floating panel view controller.
+public enum FloatingPanelState {
```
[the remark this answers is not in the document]
mhdostal: What would happen when the state was set to `.off`?

esreli:
Quick note, please rebase this PR onto the new branch `v.next-1.3`.

mhdostal:
> Quick note, please rebase this PR onto the new branch `v.next-1.3`.

@esreli Done.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+    
+    @IBOutlet private var panGestureRecognizer: UIPanGestureRecognizer!
+    
+    /// The delegate to be notified of FloatingPanelViewControllerDelegate events.
+    public var delegate: FloatingPanelViewControllerDelegate?
+    
+    /// The vertical size of the floating panel view controller used
+    /// when the `FloatingPanelState` is set to `.partial`.
+    /// The computed default is 40% of the maximum panel height.
+    public var partialHeight: CGFloat = 0
```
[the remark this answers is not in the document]
mhdostal: I kind of took the first approach as that code was rearchitected a bit. Let me know what you think.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+    
+    /// The minimum height of the floating panel, taking into account the
+    /// horizontal size class.
+    private var minimumHeight: CGFloat {
+        let headerBottom = headerStackView.frame.origin.y + headerStackView.frame.size.height
+        let handlerbarHeight = bottomHandlebarView.frame.height
+        
+        // For compactWidth, handlebar is on top, so headerBottom is the limit;
+        // For regularWidth, handlebar is on bottom, so we need to add that.
+        return isCompactWidth ? headerBottom : headerBottom + (allowManualResize ? handlerbarHeight : 0)
```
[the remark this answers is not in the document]
mhdostal: Who doesn't like a ternary within a ternary?  ;-)

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+    
+    /// Constraints for the regular width size class.
+    private var regularWidthConstraints = [NSLayoutConstraint]()
+    
+    /// Constraints for the compact width size class.
+    private var compactWidthConstraints = [NSLayoutConstraint]()
+    
+    override public func viewDidLoad() {
+        super.viewDidLoad()
+        view.translatesAutoresizingMaskIntoConstraints = false
```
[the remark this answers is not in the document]
mhdostal: I moved it to the `setupConstraints()` method.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
mhdostal: We still want the user to be able to set a `partialHeight`, based on their own content, along with the min/max heights.  Also, I've never been able to get intrinsic content sizes for TableViews, which is a lot of what's going to be displayed.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
mhdostal: That works for me.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
mhdostal: The `initalViewController` is now embedded in a `UINavigationController`, so it could end up pushing new VCs, so the FloatingPanel could end up showing multiple VCs, hence `initialViewController`.

`UINavigationController` uses `rootViewController, so I could use that if you prefer that.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.storyboard
mhdostal: Most of the stuff in the OSA apps is implemented in Storyboards, so I just kept that pattern going.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
mhdostal: Because the user can modify the size, and therefore that constraint, the source of truth is kind of both.  I did update the code to avoid unnecessarily updating the constraint constant value except when the state changes (in the new `stateDidChange` method).

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelViewController.swift
```diff
@@ -0,0 +1,484 @@
...
+            animateResizableConstrant(newConstant)
+        }
+        else {
+            resizeableLayoutConstraint.constant = newConstant
+        }
+    }
+    
+    /// The state representing the vertical size of the floating panel.
+    /// Defaults to `.partial`.
+    public var state: FloatingPanelState = .partial {
```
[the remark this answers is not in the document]
mhdostal: Maybe `state` isn't the best term, but it describes the height of the floating panel:  minimized, partial, and maximum/full.  So that bit is user-settable so that the client can decide on what the initial size should be.

mhdostal:
@esreli @philium - I've made the changes you recommended.  I've made some pretty significant changes to the architecture, including:

- rename main classes to change `ViewController` to `Controller`, similar to how subclasses of `UIViewController` like `UINavigationController`.
- rethought the `FloatingPanelItem`, now it is it's own class (not a protocol) and will be available to any `UIViewController`, which implements `FloatingPanelEmbeddable`.  Again, this is similar to `UINavigationController.navigationItem`.
- rethought the method behind the header display; now it is it's own `UINavigationController` whose push/pops are controlled by the main content navigation controller (aptly titled `contentViewController`).  Added custom animation to both view controllers, via a custom subclass of `UINavigationController`: `FloatingPanelNavigationController`.
- Added a `static func instantiate(_ initialViewController: FloatingPanelEmbeddable) -> FloatingPanelController` to simplify creating a floating panel and setting the initial VC.

I may have missed some small bits from the original reviews, so please re-raise those issues if necessary.

mhdostal:
@sbaskaran - Can you please review this along with Phil and Eli?  Thanks.

philium (commented):
Wow, you made a lot of changes! I looked through them and made some more comments and suggestions.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    
+    /// Instantiates a `FloatingPanelController` from the storyboard.
+    /// - Returns: A `FloatingPanelController`.
+    static func instantiate(_ initialViewController: FloatingPanelEmbeddable) -> FloatingPanelController {
+        // Get the storyboard for the FloatingPanelController.
+        let storyboard = UIStoryboard(name: "FloatingPanelController", bundle: .main)
+        // Instantiate the FloatingPanelController.
+        let floatingPanel = storyboard.instantiateInitialViewController() as! FloatingPanelController
+        floatingPanel.initialViewController = initialViewController
+        return floatingPanel
```
sbaskaran, line 418:
```suggestion
        let floatingPanelController = storyboard.instantiateInitialViewController() as! FloatingPanelController
        floatingPanelController.initialViewController = initialViewController
        return floatingPanelController
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+        }
+        else {
+            // On cancellation, animate the panel height to its original size.
+            animateResizableConstrant(initialResizableLayoutConstraintConstant)
+        }
+    }
+    
+    /// Updates the handlebar visibility based on horizontal size class
+    /// and the value of `allowManualResize`.
+    private func updateHandlebarVisibility() {
```
sbaskaran, line 515: Can this method be moved up, near `updateAllowManualResize()`?

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+        }
+        else {
+            animatedTransitioning = operation == .pop ? PopTransitionAnimation() : PushTransitionAnimation()
+        }
+        return animatedTransitioning//operation == .pop ? PopTransitionAnimation() : PushTransitionAnimation()
+    }
+    
+    public func navigationController(_ navigationController: UINavigationController, didShow viewController: UIViewController, animated: Bool) {
+        // If we're showing the header navigation controller, set
+        // the constraints to property size the displayed view controller.
```
sbaskaran, line 730:
```suggestion
        // the constraints to properly size the displayed view controller.
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+            headerHiddenObservation = currentFloatingPanelItem.observe(\.headerHidden, options: [.new]) { [weak self] (_, change) in
+                DispatchQueue.main.async {
+                    guard let self = self else { return }
+                    self.headerViewHidden = self.currentFloatingPanelItem.headerHidden
+                }
+            }
+            
+            allowManualResizeObservation = currentFloatingPanelItem.observe(\.allowManualResize, options: [.new]) { [weak self] (_, change) in
+                DispatchQueue.main.async {
+                    guard let self = self else { return }
```
sbaskaran, line 232: The guard statement is present in some places, and not in others (in this VC and HeaderVC files). Would be nice if they were consistent.

docs/README.md
```diff
@@ -1209,6 +1209,10 @@ These custom views are built and their layouts are managed in a storyboard.
 
 A `StyledFirstResponderLabel` is a custom `UILabel` subclass that converts the label into a first responder and styles the label upon user interaction. Similarly, `StyledTextView` and `StyledTextField` are subclasses that style (`UITextView` and `UITextField`, respectively)  to appear editable or not based on if user interaction is enabled.
 
+**Floating Panel Controller**
+
+A `FloatingPanelController` is a custom `UIViewController` subclass that allows for display of content in a "floating panel", sometimes also referred to as a "bottom sheet".  Displayed content is encapsulated in a `UIViewController` that implements the `FloatingPanelEmbeddable` protocol.  Currently both "Layers" and "Bookmarks" display in a `FloatingPanelController`.
```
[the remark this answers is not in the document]
sbaskaran: > Currently both "Layers" and "Bookmarks" display in a `FloatingPanelController`.

displayed could mean drawn, which isn't the case especially for "Layers".

sbaskaran (changes requested):

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+        return height
+    }
+    
+    /// The maximum height of the floating panel, taking into account the edge insets.
+    private var maximumHeight: CGFloat {
+        return view.superview.map { $0.frame.inset(by: internalEdgeInsets).height } ?? 320
+    }
+    
+    /// The width of the floating panel for regular width size class scenarios.
+    internal let floatingPanelWidth: CGFloat = 320
```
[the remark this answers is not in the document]
esreli: Can we make this property static? It seems like it only needs to be stored in the static scope rather than on an instance.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    private var resizeableLayoutConstraint: NSLayoutConstraint!
+    
+    /// The `resizableLayoutConstraint` constant at the beginning of
+    /// user interaction; used to reset the constant in the event the
+    /// user cancels the gesture.
+    private var initialResizableLayoutConstraintConstant: CGFloat = 0.0
+    
+    /// The insets from the edge of the screen for the regular width size class.
+    /// Defaults to `8.0, 8.0, 8.0, 8.0`
+    public var regularWidthInsets = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0) {
```
esreli, line 303: Can this be made into a static property? It doesn't appear to change.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+        didSet {
+            setupConstraints()
+        }
+    }
+    
+    /// The insets from the edge of the screen for the compact width size class.
+    /// Defaults to `8.0, 0.0, 0.0, 0.0`
+    public var compactWidthInsets = UIEdgeInsets(top: 8.0, left: 0.0, bottom: 0.0, right: 0.0) {
```
esreli, line 311: Can this be made into a static property? It doesn't appear to change.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+        didSet {
```
esreli, line 312: Because this property never changes, I don't think the `didSet { }` closure is ever called.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    
+    /// The `resizableLayoutConstraint` constant at the beginning of
+    /// user interaction; used to reset the constant in the event the
+    /// user cancels the gesture.
+    private var initialResizableLayoutConstraintConstant: CGFloat = 0.0
+    
+    /// The insets from the edge of the screen for the regular width size class.
+    /// Defaults to `8.0, 8.0, 8.0, 8.0`
+    public var regularWidthInsets = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0) {
+        didSet {
```
esreli, line 304: Because this property never changes, I don't think the `didSet { }` closure is ever called.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+        return headerViewController
+    }
+    
+    /// The constraint used by the gesture to resize the floating panel view.
+    private var resizeableLayoutConstraint: NSLayoutConstraint!
+    
+    /// The `resizableLayoutConstraint` constant at the beginning of
+    /// user interaction; used to reset the constant in the event the
+    /// user cancels the gesture.
+    private var initialResizableLayoutConstraintConstant: CGFloat = 0.0
```
esreli, line 299: If this is a constant, can it be made static?

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    }
+    
+    /// The edge insets for the current layout.
+    private var internalEdgeInsets: UIEdgeInsets {
+        return isCompactWidth ? compactWidthInsets : regularWidthInsets
+    }
+    
+    /// Denotes whether the floating panel view controller is being displayed in
+    /// a compact width layout.
+    private var isCompactWidth: Bool {
```
esreli, line 324: nice!

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    private let contentNavigationController = FloatingPanelNavigationController()
+    private let headerNavigationController = UINavigationController()
+    
+    /// Add our internal header navigation controller to the `headerView`.
+    private func setupHeaderNavigationController() {
+        headerNavigationController.view.translatesAutoresizingMaskIntoConstraints = false
+        headerNavigationController.view.backgroundColor = .gray
+        addChild(headerNavigationController)
+        headerView.addSubview(headerNavigationController.view)
+        headerNavigationController.didMove(toParent: self)
```
esreli, line 363: IIRC, this should be the last method called? It then informs the child vc that the view is now presented. I believe this   is the catalyst for the child's `viewDidAppear` method to be called.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    
+    /// Constraints for the header.
+    private var headerConstraints = [NSLayoutConstraint]()
+    
+    private let contentNavigationController = FloatingPanelNavigationController()
+    private let headerNavigationController = UINavigationController()
+    
+    /// Add our internal header navigation controller to the `headerView`.
+    private func setupHeaderNavigationController() {
+        headerNavigationController.view.translatesAutoresizingMaskIntoConstraints = false
```
esreli, line 359: Somewhere in this method we should probably call [`willMove(toParent:)`](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621381-willmove). According to the [`UIViewController` docs](https://developer.apple.com/documentation/uikit/uiviewcontroller) (see section **Implementing a Container View Controller**):

> Here are the essential methods you might need to call:
> 
>     addChild(_:)
>     removeFromParent()
>     willMove(toParent:)
>     didMove(toParent:)

In the same way that (I believe) `didMove(toParent:)` triggers `viewDidAppear`, I believe `willMove(toParent:)` triggers `viewWillAppear`.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+        contentNavigationController.delegate = self
+    }
+    
+    override public func viewDidLoad() {
+        super.viewDidLoad()
+        setupContentNavigationController()
+        setupHeaderNavigationController()
+        
+        contentNavigationController.headerNavigationController = headerNavigationController
+        contentNavigationController.createHeaderViewControllerHandler = { [weak self] (floatingPanelEmbeddable: FloatingPanelEmbeddable) in
```
esreli, line 396: You could consider making this a private method extension on `FloatingPanelEmbeddable`, just a thought.

esreli (commented):
@mhdostal i've really enjoyed reviewing your work. As an end-user, the tool is fun & intuitive to use. As a developer, the tool feels like an Apple UIKit API. The quality of this work truly cannot be understated. Bravo!

I've included a number of suggestions in code and other general suggestions below. Below are a number of high level change requests, some i've assigned to them a low priority- this means that if the effort is larger than 'small', we should attend to the request at another time.

### Floating Panel Item Close Button

As of iOS 13 UIButton supports the [`.close`](https://developer.apple.com/documentation/uikit/uibutton/buttontype/close) button type, which in effect looks like the close button you've introduced (x-close). Would you provide conditional support for devices running iOS 12 to use your (x-close) button and iOS 13 and on to use the close button supplied by UIKit?

### Rounded Bottom Corners

When the floating panel is presented for non-compact environments, the nw & ne corners are rounded but the sw & se corners are not. In this environment, can you round the bottom corners as well?

Screen Shot 2020-09-24 at 11 29 21 AM

### Layers Item Icon

There is a calcite team designed icon named "layers" that is used in the [.Net app](https://github.com/Esri/data-collection-dotnet/tree/dev
). I think this icon more accurately depicts the concept of layers. Because of this and for app narrative parity, can you swap the "layer" icon found in this app with the "layers" icon found in the .Net build? 

| iOS | .Net |
| --- | --- |
| Screen Shot 2020-09-24 at 11 31 01 AM | Screen Shot 2020-09-23 at 4 51 11 PM |

### Floating Panel Item Spacing

It appears the spacing between the icon and title is 4pt. To me, this feels a bit cramped. Can we bump this up to 8pt?

Screen Shot 2020-09-24 at 11 35 37 AM

### Floating Panel Header Controller Storyboard

_This is not a requirement but a suggestion._

At first glance of the `FloatingPanelHeaderController` storyboard, it doesn't make much sense. You can modify the "Simulated Size" to "Freeform" to simulate a height that more closely (or exactly) resembles the height of the view controller. This allows the storyboard to appear to look more closely like it would as it's used by the tool. I took a stab at this below...

Screen Shot 2020-09-24 at 11 52 29 AM

### Adjusted Content Inset

_This is a low priority comment._

Perhaps i've overlooked this, does this implementation adjust the [`AGSMapView.adjustedContentInset`](https://developers.arcgis.com/ios/latest/api-reference/interface_a_g_s_map_view_common.html#abf8790df308c51d07e98d0d87e220741)? This might be something to consider for this PR if you think it's an easily implemented feature adjustment. If not, we can attend to this down the road. Thoughts?

### Attribution Bar

_This is a low priority comment._

A requirement for publishing an app built with Runtime is to not obfuscate the attribution bar, though admittedly this requirement is [vaguely defined](https://github.com/ArcGIS/open-source-apps/issues/95#issuecomment-483720590). When the app is in landscape orientation or running on an iPad, is it possible for the bottom of the floating panel to stop at the top of the attribution bar? FWIW, I think it's okay to obfuscate the attribution bar in iPhone portrait orientation because it covers the bar entirely whereas in the other presentation configuration, it obfuscates the bar only where the floating panel extends. Does this make sense?

Screen Shot 2020-09-24 at 11 28 27 AM

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
mhdostal: It can be set by clients to adjust the location of the floating panel.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
mhdostal: It can be set by clients to adjust the location of the floating panel.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
mhdostal: See above...

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.storyboard
```diff
@@ -0,0 +1,260 @@
...
+                                                </userDefinedRuntimeAttributes>
+                                            </view>
+                                        </subviews>
+                                        <constraints>
+                                            <constraint firstItem="5d5-9O-fRp" firstAttribute="centerY" secondItem="63a-iz-mY9" secondAttribute="centerY" id="b8I-Tj-ShI"/>
+                                            <constraint firstItem="5d5-9O-fRp" firstAttribute="centerX" secondItem="63a-iz-mY9" secondAttribute="centerX" id="hd9-tv-ZO9"/>
+                                            <constraint firstAttribute="height" constant="16" id="uvX-dm-aSJ"/>
+                                        </constraints>
+                                    </view>
+                                    <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="Thq-YG-pDT" userLabel="Header View">
```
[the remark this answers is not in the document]
mhdostal: Leftover from debugging, nice catch.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.storyboard
```diff
@@ -0,0 +1,260 @@
...
+                            <stackView opaque="NO" contentMode="scaleToFill" insetsLayoutMarginsFromSafeArea="NO" axis="vertical" translatesAutoresizingMaskIntoConstraints="NO" id="AwH-UW-fqK" userLabel="Header Stack View">
+                                <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+                                <subviews>
+                                    <view contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="t1Z-3F-jll" userLabel="Header Spacer View">
+                                        <rect key="frame" x="0.0" y="0.0" width="414" height="16"/>
+                                        <constraints>
+                                            <constraint firstAttribute="height" constant="16" id="aud-TR-FqT"/>
+                                        </constraints>
+                                    </view>
+                                    <stackView opaque="NO" contentMode="scaleToFill" translatesAutoresizingMaskIntoConstraints="NO" id="d7h-Gh-ygK" userLabel="Header Detail Stack View">
```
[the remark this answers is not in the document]
mhdostal: Excellent, thank makes things much cleaner!

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    dynamic var closeButtonHidden: Bool = false
+    
+    /// The vertical size of the floating panel view controller used
+    /// when the `FloatingPanelController.state` is set to `.partial`.
+    @objc
+    dynamic var partialHeight: CGFloat = FloatingPanelController.automaticDimension
+    
+    /// The state representing the vertical size of the floating panel.
+    @objc
+    dynamic var state: FloatingPanelController.State = .partial
```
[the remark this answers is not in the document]
mhdostal: How about `verticalState`, since the doc says "...representing the vertical size.."?

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
esreli: Good call 👍

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+        case minimized
+        /// The floating panel is displayed at a height roughly 40% of the screen.
+        /// The actual value is user-customizable.
+        case partial
+        /// The floating panel is displayed at its maximum height.
+        case full
+    }
+
+    @IBOutlet private var contentView: UIView!
+    @IBOutlet internal var headerView: UIView!
```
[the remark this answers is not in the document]
mhdostal: I changed it to `fileprivate`.  The class is marked as `public` but I want to keep the access to `headerView` not public.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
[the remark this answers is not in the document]
mhdostal: made it `static`.  Thanks!

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    
+    /// The width of the floating panel for regular width size class scenarios.
+    internal let floatingPanelWidth: CGFloat = 320
+    
+    /// The vertical state of the floating panel.  When changed, the
+    /// floating panel will animate to the new state.
+    /// - Parameter animated: Denotes whether to animate to the new view height
+    /// resulting from the change in state.
+    private func stateDidChange(animated: Bool = true) {
+        let newConstant: CGFloat
```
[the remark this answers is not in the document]
mhdostal: `updatedConstraintConstant`

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+            animateResizableConstrant(newConstant)
+        }
+        else {
+            resizeableLayoutConstraint.constant = newConstant
+        }
+    }
+    
+    private func updateHeaderViewVisibility() {
+        headerView.isHidden = headerViewHidden
+        headerView.alpha = headerViewHidden ? 0 : 1
```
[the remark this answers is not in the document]
mhdostal: No.  Just setting `isHidden` now.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,822 @@
...
+    }
+    
+    /// Denotes whether the floating panel view controller is being displayed in
+    /// a compact width layout.
+    private var isCompactWidth: Bool {
+        return traitCollection.horizontalSizeClass == .compact
+    }
+    
+    /// Determines whether we've initialized the initial layout constraints.
+    private var initialized = false
```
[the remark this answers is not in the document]
mhdostal: Changed to `layoutConstraintsInitialized`.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
mhdostal: The rule of thumb I use is that if `self?.` is used in more than one place, then I'll add the `guard` statement.  Otherwise it's just easier to use `self?.`  I did make a change or two to enforce my own rule... ;-)

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
mhdostal: It can be set by clients.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
mhdostal: It's actually set by the `handlePanGesture` method in the `.began` case of the switch.  It's used to reset the state based on the outcome of the pan gesture.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
mhdostal: I dug into this and calling `addChild` will automatically call `willMove`:

```When your custom container calls the addChildViewController: method, it automatically calls the willMoveToParentViewController: method of the view controller to be added as a child before adding it.```

**Comment by mhdostal** [comment]
> As of iOS 13 UIButton supports the [`.close`](https://developer.apple.com/documentation/uikit/uibutton/buttontype/close) button type, which in effect looks like the close button you've introduced (x-close). Would you provide conditional support for devices running iOS 12 to use your (x-close) button and iOS 13 and on to use the close button supplied by UIKit?

@esreli - the `buttonType` property on `UIButton` is `readonly`, so I have to create the button with that type, either in the Storyboard or programmatically.  Is there any mechanism to have such conditional-construction in the Storyboard?

**Comment by esreli** [comment]
https://github.com/Esri/data-collection-ios/pull/247#issuecomment-699097074

@mhdostal, I don't believe there is. This would probably require building the `UIButton` programmatically.

File: data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift (hunk 38 above)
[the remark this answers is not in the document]
Comment (sbaskaran) [reply data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift:42]: Sounds good

File: data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift [hunk 43]
```diff
@@ -0,0 +1,822 @@
...
+    
+    override func popViewController(animated: Bool) -> UIViewController? {
+        let returnVC = super.popViewController(animated: animated)
+        headerNavigationController?.popViewController(animated: animated)
+        
+        return returnVC
+    }
+}
+
+extension UIViewController {
```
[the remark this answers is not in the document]
mhdostal: I'm not sure I see the benefit of the `FloatingPanelPresenter` protocol.  Whatever implements that protocol will still have to implement the `floatingPanel` property.  And if the protocol is implemented by the view controller presenting the panel, I don't see a use case for an external object to get the `floatingPanel` property of the `FloatingPanelPresenter`; that seems like something internal to the controlling object.

Having the `present/dismissFloatingPanel` methods on an extension of `UIViewController` allows all view controllers to present/dismiss a floating panel.

I did redo them slightly, based on your comment above regarding `AGSMapView. adjustedContentInset`.  The creation of the floating panel is now left to the client, who would then use the `presentFloatingPanel` extension method to display it.  Since the constraints are created in the `UIViewController` extension, there is no facility to re-create them if the user changes the edge inset properties.  Therefore, I have added the two insets as arguments to the `FloatingPanelController.instantiate(...)` method as follows:

```
    static func instantiate(_ initialViewController: FloatingPanelEmbeddable,
                            regularWidthInsets: UIEdgeInsets = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0),
                            compactWidthInsets: UIEdgeInsets = UIEdgeInsets(top: 8.0, left: 0.0, bottom: 0.0, right: 0.0)) -> FloatingPanelController {
```
And I've made the two inset properties read-only.  With that constructor, the user can pass in the `adjustedContentInsets` to have the floating panel not interfere with the attribution bar.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
mhdostal: The implementation of that handler calls a private method on FloatingPanelController, so I don't think that would work.

mhdostal:
> [#247 (comment)](https://github.com/Esri/data-collection-ios/pull/247#issuecomment-699097074)
> 
> @mhdostal, I don't believe there is. This would probably require building the `UIButton` programmatically.

Ok.  I'm going to leave it as-is then and we can re-look at it when the SDK moves to iOS 13 as a minimum.

mhdostal:
@esreli, @philium , @sbaskaran - I've address all the feedback, please take another look.  Thanks!

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,792 @@
...
+        updateHandlebarVisibility()
+        updateInterfaceForCurrentTraits()
+        
+        // Set corner masks
+        var maskedCorners: CACornerMask = [.layerMinXMinYCorner, .layerMaxXMinYCorner]
+        if !isCompactWidth {
+            maskedCorners.formUnion([.layerMaxXMaxYCorner, .layerMinXMaxYCorner])
+        }
+        view.layer.maskedCorners = maskedCorners
+        view.clipsToBounds = true
```
philium, line 538: Since this is always set to the same value, a more appropriate place for this would be `viewDidLoad()` so that it is only set once.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,792 @@
...
+                       animations: {
+                        self.resizeableLayoutConstraint.constant = newConstant
+                        self.contentView?.alpha = (self.currentFloatingPanelItem.verticalState == .minimized ? 0.0 : 1.0)
+                        self.view.setNeedsUpdateConstraints()
+                        superview.layoutIfNeeded()
+        })
+    }
+    
+    /// For some reason, the first time the initial view controller is set there
+    /// are two calls to `navigationController:didShow:viewController:animated`.
```
philium, line 564:
```suggestion
    /// are two calls to `navigationController(_:didShow:animated:)`.
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,792 @@
...
+    /// The second call messes up the logic for which header view controller to show.
+    /// Adding a check for `animating` prevents the second call from happening.
+    fileprivate var animating = false
+}
+
+fileprivate let transitionAnimationDuration: TimeInterval = 0.5
```
philium, line 570: At the top level, `fileprivate` is equivalent to `private`. A good rule of thumb is this: when needing to make something private, default to `private` and only use `fileprivate` when absolutely necessary.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,792 @@
...
+                view.safeAreaLayoutGuide.bottomAnchor.constraint(greaterThanOrEqualTo: floatingPanel.headerView.bottomAnchor, constant: floatingPanel.compactWidthInsets.bottom)
+            ]
+        }
+    }
+    
+    /// Dismisses the given `FloatingPanelController`.
+    func dismissFloatingPanel(_ floatingPanel: FloatingPanelController) {
+        floatingPanel.willMove(toParent: nil)
+        floatingPanel.removeFromParent()
+        floatingPanel.view.removeFromSuperview()
```
philium, line 790: See [Implementing a Container View Controller](https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html).
```suggestion
        floatingPanel.willMove(toParent: nil)
        floatingPanel.view.removeFromSuperview()
        floatingPanel.removeFromParent()
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelHeaderController.swift
```diff
@@ -0,0 +1,117 @@
...
+internal class FloatingPanelHeaderController: UIViewController {
+    @IBOutlet private var closeButton: UIButton!
+    @IBOutlet private var titleLabel: UILabel!
+    @IBOutlet private var subtitleLabel: UILabel!
+    @IBOutlet private var imageView: UIImageView!
+    @IBOutlet private var subtitleSpacerView: UIView!
+    
+    /// Observers for `FloatingPanelItem` properties that are
+    /// displayed by the `FloatingPanelHeaderController`.
+    private var closeButtonObservation: NSKeyValueObservation?
```
philium, line 27: FYI: You could just put these observations into an array and the empty the array before creating new observations. That would simplify adding more observations in the future (if needed) as a new property wouldn't be needed. Just a thought.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelHeaderController.swift
```diff
@@ -0,0 +1,117 @@
...
+//        if #available(iOS 13.0, *) {
+//            // In iOS 12.x, this used to crash without using `.alert` or setting either
+//            // a UIBarButtonItem or sourceView and rect.  Fixed in iOS 13.
+//            closeButton.buttonType = .close
+//        }
+//        else {
+//            style = .alert
+//        }
+//
+    }
```
philium, line 95:
```suggestion
```

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -145,3 +124,48 @@ extension MapViewController: BookmarksViewControllerDelegate {
...
+        get {
+            let fpItem = FloatingPanelItem()
+            fpItem.title = MapViewControllerExtras.bookmarks.title
+            fpItem.subtitle = "Select a bookmark"
+            fpItem.image = UIImage(named: "bookmark")
+            return fpItem
+        }
+        set {
+            //No-op
+        }
```
philium, line 154:
```suggestion
        let fpItem = FloatingPanelItem()
        fpItem.title = MapViewControllerExtras.bookmarks.title
        fpItem.subtitle = "Select a bookmark"
        fpItem.image = UIImage(named: "bookmark")
        return fpItem
```

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+Extras.swift
```diff
@@ -145,3 +124,48 @@ extension MapViewController: BookmarksViewControllerDelegate {
...
+    // changes to the floating panel item, so it's OK to just have
+    // this as a computed property.  Otherwise, that class would need
+    // to be updated to implement FloatingPanelEmbeddable.
+    public var floatingPanelItem: FloatingPanelItem {
+        get {
+            let item = FloatingPanelItem()
+            item.title = MapViewControllerExtras.layers.title
+            item.image = UIImage(named: "layers")
+            return item
+        }
```
philium, line 169:
```suggestion
        let item = FloatingPanelItem()
        item.title = MapViewControllerExtras.layers.title
        item.image = UIImage(named: "layers")
        return item
```

philium (commented):
Looking better! I've re-reviewed and have left a few more suggestions.

mhdostal:
@philium I made the requested changes, ready for re-review.  Thanks!

philium (approved):

sbaskaran (approved):

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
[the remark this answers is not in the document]
esreli: > I'm not sure I see the benefit of the `FloatingPanelPresenter` protocol.

At bare minimum, it would allow you to remove some reused code and reduce some of the burden placed on the controlling object. For instance, implementing `presentFloatingPanel` as an extension:

```swift
protocol FloatingPanelPresenter: class {
    var floatingPanelController: FloatingPanelController? { get set }
}

extension FloatingPanelPresenter where Self: FloatingPanelControllerDelegate, Self: UIViewController {
    
    func presentInFloatingPanel(_ embeddable: FloatingPanelEmbeddable, insets: UIEdgeInsets) {
        let floatingPanel = FloatingPanelController.instantiate(embeddable, regularWidthInsets: insets)
        floatingPanelController = floatingPanel
        presentFloatingPanel(floatingPanel)
        floatingPanel.delegate = self
    }
}
```

`MapViewController` would then adhere to `FloatingPanelPresenter`, giving the controller floating panel presentation/dismissing privileges.

```swift

extension MapViewController: FloatingPanelPresenter { }

extension MapViewController {

    func showBookmarks(_ barButtonItem: UIBarButtonItem?) {
        // Configure the view controller.
        let bookmarksVC = BookmarksViewController(geoView: mapView)
        bookmarksVC.delegate = self

        // Use the new extension method
        presentInFloatingPanel(bookmarksVC, insets: adjustedFloatingPanelInsets())
    }
}
```

You'll see here, this reduces the burden in `showBookmarks`, `showLayerContents`, and any other case written in the future where `MapViewController` wishes to present a `FloatingPanelPresentable` in the floating panel.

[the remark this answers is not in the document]
mhdostal: Ok, I get it.  I've made the change.  It does make the client code easier to write/maintain.

esreli (approved):
All of my feedback has been addressed. I think this is a wonderful tool and I can't wait to see it used even more broadly. Great work 👍

rolson:
👍

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,812 @@
...
+    override func popViewController(animated: Bool) -> UIViewController? {
+        let returnVC = super.popViewController(animated: animated)
+        headerNavigationController?.popViewController(animated: animated)
+        
+        return returnVC
+    }
+}
+
+/// Protocol designed to be implemented by objects that will display a floating panel.
+protocol FloatingPanelPresenter: class {
```
philium, line 761: Since this protocol is really only useful for subclasses of `UIViewController`, it might be better to just express that as a constraint of the protocol itself:

```swift
protocol FloatingPanelPresenter: UIViewController {
```

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,812 @@
...
+    /// The floating panel to be displayed.
+    var floatingPanelController: FloatingPanelController? { get set }
```
philium, line 763: Instead of making this a requirement of the protocol, it could just be computed, assuming the protocol is constrained to subclasses of `UIViewController`:

```swift
extension FloatingPanelPresenter {
    var floatingPanelController: FloatingPanelController? {
        children.first(where: { $0 is FloatingPanelController }) as? FloatingPanelController
    }
}
```

That would make the protocol more of a mix-in, providing additional behavior that a view controller would obtain by declaring conformance to it.

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -0,0 +1,812 @@
...
+    ///   - regularWidthInsets: The UIEdgeInsets used in a regular width horizontal size class.
+    ///   - compactWidthInsets: The UIEdgeInsets used in a compact width horizontal size class.
+    func presentInFloatingPanel(_ embeddable: FloatingPanelEmbeddable,
+                                regularWidthInsets: UIEdgeInsets = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0),
+                                compactWidthInsets: UIEdgeInsets = UIEdgeInsets(top: 8.0, left: 0.0, bottom: 0.0, right: 0.0)) {
+        let floatingPanel = FloatingPanelController.instantiate(embeddable,
+                                                                regularWidthInsets: regularWidthInsets,
+                                                                compactWidthInsets: compactWidthInsets)
+        floatingPanelController = floatingPanel
+        floatingPanel.delegate = self
```
philium, line 779: Setting the delegate is just a convenience that, were it to be absent, wouldn't hinder the presentation of a floating panel. So perhaps instead of constraining the extension to types that conform to `FloatingPanelControllerDelegate`, this line could just do a cast:

```swift
floatingPanel.delegate = self as? FloatingPanelControllerDelegate
```

PR #248, opened 2020-08-04
Use credentialCache.removeAndRevokeAllCredentials when signing out.

mhdostal:
Data Collection to support clearing and revoking credential -   As of 100.6, Runtime introduced a method to [clear and revoke credentials](https://developers.arcgis.com/ios/latest/api-reference/interface_a_g_s_credential_cache.html#a0796cf2506fa0edfdeb2b62198bbbea7). This introduction came after the initial general release of Data Collection. Calling this method revokes the portal user's credential on the server side.

Let's retroactively revisit Data Collection to ensure this method is called when a user requests to log out. 

Additional considerations should be made to ensure that services that have been supplied a previously valid credential will issue a challenge again after that credential is revoked.

### DIFF
--- a/data-collection/data-collection/App Context/AppContext+Login.swift
+++ b/data-collection/data-collection/App Context/AppContext+Login.swift
@@ -52,8 +52,16 @@ extension AppContext {
     ///
     /// The app does this by removing all cached credentials and no longer requiring authentication in the portal.
     func signOut() {
-        // We want to remove cached credentials upon sign-out.
-        AGSAuthenticationManager.shared().credentialCache.removeAllCredentials()
+        // We want to remove cached credentials and revoke tokens upon sign-out.
+        AGSAuthenticationManager.shared().credentialCache.removeAndRevokeAllCredentials { (results) in
+            if !results.isEmpty {
+                print("Could not revoke the following credentials:")
+                results.forEach { (credential, error) in
+                    print("  \(credential.username ?? ""); error: \(error.localizedDescription)")
+                }
+            }
+        }
+        
         // We want to remove cached credentials from geo-coder services, in case they are cached.
         appAddressLocator.removeCredentialsFromServices()
         // Setting `loginRequired` to `false` will allow unauthenticated users to consume the map (but not edit!)


esreli:
Quick note here, please rebase onto branch `v.next-1.2.2`

mhdostal:
> Are we sure the order of operations are correct here?
> 
> The first method performed clears credentials locally and attempts to revoked them on the server. This method is async. Should we perform `appAddressLocator.removeCredentialsFromServices()` in the callback? Should we set the new portal `portal = AGSPortal.configuredPortal(loginRequired: false)` in the callback?
> 
> I'm not exactly sure, though that does seem correct... at least to me.

I don't think it matters.  The reason the `removeCredentialsFromServices()` is async is it's going over the network.  That may/may not take a while and I think it's important for the workflow that once the user "signs out", the UI updates quickly, which *wouldn't* happen if we did it in the async handler.  The operations are independent of each other, so I don't think they need to move to the async completion handler.

mhdostal:
@esreli How about we move the sign-in/out and potential `PortalSessionManager` updates/fixes to a future release?  That way we can get this out the door, with the 100.9 Toolkit update and 100.9.0 SDK certification.  The proposed portal sign-in/out changes are more involved.

esreli:
> The operations are independent of each other, so I don't think they need to move to the async completion handler.
>
> @mhdostal 

Sounds good! Last consideration, what if the user is disconnected? Would it be that the server isn't aware the user is trying to sign out but because the credentials are wiped locally, it doesn't matter if the server persists them because they will never again be used? Just trying to spitball some different cases here...

> @esreli How about we move the sign-in/out and potential `PortalSessionManager` updates/fixes to a future release?
>
> @mhdostal 

Sounds good, i've created an [issue](https://github.com/ArcGIS/open-source-apps/issues/444) to capture this work down the road.

mhdostal:
FYI, here is the test code I used:

This produces a “Credential Required” error the second time `fetchContent` is called:

```
fetchOp = portal.user?.fetchContent(completion: { [weak self] (items, folders, error) in
    print("fetchContent error = \(String(describing: error?.localizedDescription))")

    AGSAuthenticationManager.shared().credentialCache.removeAndRevokeAllCredentials { [weak self] (results) in
        if !results.isEmpty {
            print("Could not revoke the following credentials:")
            results.forEach { (credential, error) in
                print("  \(credential.username ?? ""); error: \(error.localizedDescription)")
            }
        }

        //Test and make sure credentials are revoked
        self?.fetchOp = self?.portal.user?.fetchContent(completion: { (items, folders, error) in
            print("post-signOut fetchContent error = \(String(describing: error?.localizedDescription))")
        })
    }
})
```

The output I get is:

```
fetchContent error = nil
post-signOut fetchContent error = Optional("Credential Required")
```

mhdostal:
> Sounds good! Last consideration, what if the user is disconnected? Would it be that the server isn't aware the user is trying to sign out but because the credentials are wiped locally, it doesn't matter if the server persists them because they will never again be used? Just trying to spitball some different cases here...
> 
> >

@njarecha Can you answer @esreli question?  Essentially, would the behavior be the same as the pre-`credentialCache.removeAndRevokeAllCredentials` (meaning only clearing the local credential cache) behavior if the user was disconnected when `credentialCache.removeAndRevokeAllCredentials` was called?

mhdostal:
> Yes, in case of any error (including network error) while while calling `removeAndRevokeAllCredentials` behavior same as before this method. I would suggest dismissing view controller after `SignOut` and setting portal object to `nil` are new instance for everything to work correctly.

Thanks @njarecha .  We are already setting the portal to a new instance, so we should be good.

esreli (commented):
https://github.com/Esri/data-collection-ios/pull/248#issuecomment-687337561
@mhdostal it looks like we're all set here. Can you add a line to the RELEASE notes detailing this change?

PR #255, opened 2020-10-01
Mhd/floating panel 2

mhdostal:
Some tweaks and simplification of the `FloatingPanelPresenter`.

### DIFF
--- a/data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
+++ b/data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
@@ -757,8 +757,7 @@ fileprivate class FloatingPanelNavigationController: UINavigationController {
     }
 }
 
-/// A convenience protocol that can be implemented by view controllers
-/// displaying a floating panel to easily present and dismiss the panel.
+/// An extension allowing view controllers to easily present and dismiss a floating panel.
 extension UIViewController {
     /// The presented floating panel controller.  This will be nil
     /// until `presentInFloatingPanel` has been called.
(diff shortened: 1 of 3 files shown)


data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -757,13 +757,17 @@ fileprivate class FloatingPanelNavigationController: UINavigationController {
...
-/// Protocol designed to be implemented by objects that will display a floating panel.
-protocol FloatingPanelPresenter: class {
-    /// The floating panel to be displayed.
-    var floatingPanelController: FloatingPanelController? { get set }
-}
+/// A convenience protocol that can be implemented by view controllers
+/// displaying a floating panel to easily present and dismiss the panel.
+protocol FloatingPanelPresenter: UIViewController { }
+
+extension FloatingPanelPresenter {
```
esreli, line 764: I think these changes are great. With them, however, it appears the need for `FloatingPanelPresenter` has disappeared. The pattern established here is very nice and is conducive such that I think we could restore this tool for use with any `UIViewController` pretty easily.

```suggestion
extension UIViewController {
```

(If you take this suggestion, don't forget to remove the `MapViewController`'s adherence to `FloatingPanelPresenter`)

philium: @esreli 's suggestion makes sense.

mhdostal:
@esreli, @philium I've committed and merged the changes we talked about.  Thank you for all your help on this!

data-collection/data-collection/View Controllers/Floating Panel View Controller/FloatingPanelController.swift
```diff
@@ -757,8 +757,7 @@ fileprivate class FloatingPanelNavigationController: UINavigationController {
     }
 }
 
-/// A convenience protocol that can be implemented by view controllers
-/// displaying a floating panel to easily present and dismiss the panel.
+/// An extension allowing view controllers to easily present and dismiss a floating panel.
```
philium, line 760: FYI: While there isn't any harm in using a documentation comment on an extension, I don't think there's any benefit to doing it.

PR #258, opened 2020-11-11
Update paths for ArcGIS.xcframework

mhdostal:
Update paths for ArcGIS.xcframework as the next release of the ArcGIS SDK now builds a binary framework.

Includes Xcode 12 auto project updates.


data-collection/data-collection.xcodeproj/project.pbxproj
```diff
@@ -1547,7 +1531,7 @@
 				94C550F622123C2E00DD7E40 /* PopupSelectRelatedRecordCell.swift in Sources */,
 				948D85232151C11F00E2048F /* AGSPopupManager+FeatureTable.swift in Sources */,
 				94D2714020D0315900304E27 /* CustomBehavior_Map.swift in Sources */,
-				946D20C22130C236001BF5CC /* LastSyncMobileMapPackage.swift in Sources */,
+                946D20C22130C236001BF5CC /* LastSyncMobileMapPackage.swift in Sources */,
```
philium, line 1534: Please revert this change.

philium (changes requested):

esreli (commented):
Thank you, @mhdostal. This is an exciting change for Runtime on iOS! Would you please describe the change you've made here in the `RELEASE.md` notes? Additionally, you might make a note in the `README.md` doc's `#get-started` section, that this change is introduced with Runtime v 100.10? Because this project looks for a framework at the new default install location, if a customer of ours is not ready to upgrade to Runtime v 100.10, they will have to remove this new .xcframework reference in favor of the old framework's path. << This too should be noted.

mhdostal:
@esreli I have made the requested changes to the README and RELEASE files.  However, we may have to revisit the section in the README concerning use of the 100.9 SDK with the 1.2.3 release of data collection.  There may be other changes which use new methods in the 100.10 SDK, such as the new baseman styles stuff.  A number of the old basemap methods have been deprecated, and if we switch to the new methods will will require SDK v100.10.

I'd say we revisit that during the 100.10 certification.

philium:
It seems to me that just making 100.10.0 a requirement (which according to the readme it already is) is sufficient. With 100.10.0 installed, as far as the Data Collection app is concerned, the binary framework changes would just be transparent, correct? The 100.10.0 release of the SDK will include information about the binary framework changes so that users can update their own projects. Perhaps it is unnecessary to duplicate that information in each OSA project?

mhdostal:
I'm fine with requiring 100.10, especially as we're probably making changes to the baseman styles API that will force 100.10.

esreli:
> It seems to me that just making 100.10.0 a requirement (which according to the readme it already is) is sufficient.

It doesn't appear that @mhdostal has explicitly required Runtime v100.10 in this release- or Xcode 12 for that matter (am I mistaken in my understanding that Xcode 12 is now a requirement for building with Runtime 100.10?) The doc in its current state reads as follows,

> You will need [Xcode 11](https://itunes.apple.com/us/app/xcode/id497799835?mt=12) and the [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/latest/swift/guide/install.htm#ESRI_SECTION1_D57435A2BEBC4D29AFA3A4CAA722506A) (v100.8 or later) installed locally.

The requirement we've specified here is to introduce support for 100.10. This does not necessarily mean we need to remove support for pre 100.10 Runtime versions. Historically OSA has not required the latest version of Runtime with every new release and instead required only up to the latest release where a Runtime bug fix or a new feature API was introduced. I'm open to the idea of dropping support for pre-100.10 versions of Runtime if we can justify it, especially with this introduction of the binary framework, but as of now we aren't defaulting to requiring the latest version of Runtime.

philium:
> It doesn't appear that @mhdostal has explicitly required Runtime v100.10 in this release

The latest commit does specify 100.10.0 as the minimum required version of runtime:

image

> am I mistaken in my understanding that Xcode 12 is now a requirement for building with Runtime 100.10?

I believe the 100.10.0 release can still be used in projects built with Xcode 11.

esreli:
https://github.com/Esri/data-collection-ios/pull/258#issuecomment-726922131

This, then, contradicts a previous mention of requirements.

> ## Get started
>
> You will need [Xcode 11](https://itunes.apple.com/us/app/xcode/id497799835?mt=12) and the [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/latest/swift/guide/install.htm#ESRI_SECTION1_D57435A2BEBC4D29AFA3A4CAA722506A) (v100.8 or later) installed locally.
> 
> The *Data Collection app* now incorporates the [ArcGIS Runtime Toolkit for iOS](https://github.com/Esri/arcgis-runtime-toolkit-ios) (v100.10.0 or later) for additional functionality.
> 
> Note that the 100.10.0 release of the ArcGIS Runtime SDK for iOS replaces the installed "fat framework" `ArcGIS.framework` with a new binary framework `ArcGIS.xcframework`.  It also changes the location of the installed framework file and removes the need for the `strip-frameworks.sh` Build Phase.  These changes have been incorporated in the lastest release of Data Collection.  If you do not wish to upgrade to the 100.10.0 release of the SDK, but still wish to use the latest Data Collection release, you will need to remove the `ArcGIS.xcframework` embedded file, re-add the pre-100.10.0 `ArcGIS.framework` and re-add the Build Phase which executed the `strip-frameworks.sh` shell script.  You will also need to use the 100.9.0 release of the ArcGIS Toolkit for iOS.

It looks as if we should resolve these two conflicting requirements. Better yet, we could mention requirements only once.

esreli:
> I believe the 100.10.0 release can still be used in projects built with Xcode 11.

Got it, I must be confusing this with something else. (Is it now a requirement for internally building the SDK?)

mhdostal:
I have clarified the Xcode and SDK version requirements.  Xcode 11 is still OK, as the Toolkit was updated to not require Xcode 12.  Also, v100.10 SDK is required because we will be deprecating some of the base map methods in favor of the new base map styles.

mhdostal:
@esreli @philium Please re-review.

README.md
```diff
@@ -63,9 +63,9 @@ Read the [docs](./docs/README.md) for a detailed explanation of the application,
 
 ## Get started
 
-You will need [Xcode 11](https://itunes.apple.com/us/app/xcode/id497799835?mt=12) and the [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/ios/latest/swift/guide/install.htm#ESRI_SECTION1_D57435A2BEBC4D29AFA3A4CAA722506A) (v100.8 or later) installed locally.
+Make sure you've installed Xcode and the ArcGIS Runtime SDK for iOS and that they meet these [requirements](#requirements).
 
-The *Data Collection app* now incorporates the [ArcGIS Runtime Toolkit for iOS](https://github.com/Esri/arcgis-runtime-toolkit-ios) (v100.8 or later) for additional functionality.
+Note that the 100.10 release of the ArcGIS Runtime SDK for iOS replaces the installed "fat framework" `ArcGIS.framework` with a new binary framework `ArcGIS.xcframework`.  It also changes the location of the installed framework file and removes the need for the `strip-frameworks.sh` Build Phase.  These changes have been incorporated in the lastest release of the *Data Collection app*.
```
esreli, line 68: I'm wondering if this line would be better suited in the **Requirements** section. Your thoughts?
```suggestion
**Note**, the 100.10 release of the ArcGIS Runtime SDK for iOS replaces the installed "fat framework" `ArcGIS.framework` with a new binary framework `ArcGIS.xcframework`.  It also changes the location of the installed framework file and removes the need for the `strip-frameworks.sh` Build Phase.  These changes have been incorporated in the lastest release of the *Data Collection app*.
```

esreli (approved):
I've built and run the project that includes the new binary framework with no issues, doc has been resolved and (both) build phase run scripts are removed. This PR LTGM, thank you!

philium (approved):

mhdostal:
@esreli @philium Thank you both for your help with this!

PR #259, opened 2020-11-13
Moves single-use extensions from isolated file to caller file

esreli:
_Note, no code change is performed in this PR._ This change is progressively performed to clean up the project structure by moving existing single-use extensions from their isolated file to the caller file.


philium:
For those that you can (like `AGSMap`), I would suggest making the extension `private`. As for the other extensions (like those that add conformance to a protocol), they can't be made private, which is a good argument for keeping them in separate source files. Since they are non-private, they could end up being used elsewhere at any time, making them non-single-use. Just a thought.

esreli:
>  I would suggest making the extension `private`.

This is a good call.

esreli:
@mhdostal @philium i'm marking this first batch as ready for review. Would you have a look and offer your feedback? Thank you in advance.

data-collection/data-collection/App Context/OfflineMapManager.swift
```diff
@@ -390,3 +390,17 @@ fileprivate extension URL {
 protocol JobResult {}
 extension AGSGenerateOfflineMapResult: JobResult {}
 extension AGSOfflineMapSyncResult: JobResult {}
+
+
+extension AGSMap {
```
philium, line 395:
```suggestion
private extension AGSMap {
```

data-collection/data-collection/Extensions/ArcGIS/AGSArcGISFeatureTable/AGSArcGISFeatureTable+Query.swift
```diff
@@ -162,3 +162,16 @@ extension AGSArcGISFeatureTable {
         }
     }
 }
+
+// MARK: Query Parameters
+
+private extension AGSQueryParameters {
+    
+    /// Return a new `AGSQueryParameters` with a `whereClause` requesting all features in the table.
+
```
philium, line 171:
```suggestion
```

philium (approved):

philium (approved):

esreli:
@mhdostal have you had a chance to review this PR? what do you think?

PR #263, opened 2020-11-19
Drops conformance to AppErrors and reworks errors as members of related types

esreli:
This PR is attempting to:
- Migrate app errors from a global space into types from where they are provided/thrown.
- Eliminate the `AppError` protocol, a contrived NSError derived protocol that makes too many assumptions on what the end developer need from it

_Note, i'm not sure how the removal of `LastSyncMobileMapPackage.swift` from the build phase occurred, my best guess is that it is a legacy token left behind by a previous improper Xcode removal and that Xcode is attempting to self-correct._


data-collection/data-collection/Rich Popup/Related Records/Relationships.swift
```diff
@@ -117,19 +117,23 @@ class Relationships: AGSLoadableBase {
             
             self.loadingRelationships = nil
             
-            var errors = [Error]()
+            let errors: [(String, Error)] = loadedRelationships.compactMap { (relationship) in
+                if let error = relationship.loadError {
+                    return ((relationship as! Relationship).name ?? "(missing tablename)", error)
```
philium, line 122: This cast could be eliminated if the type of `Relationships.loadingRelationships` is changed from `[AGSLoadable]?` to `[Relationship]?`.

data-collection/data-collection/Rich Popup/Related Records/Relationships.swift
```diff
@@ -139,12 +143,14 @@ class Relationships: AGSLoadableBase {
...
-                self.loadDidFinishWithError(nil)
-            }
+            self.loadDidFinishWithError(nil)
+        }
+    }
+    
+    struct RelationshipsLoadError: LocalizedError {
+        let errors: [(tablename: String, error: Error)]
+        var localizedDescription: String {
+            "Failed to load relationships: \(errors.map{$0.tablename}.joined(separator: ","))"
```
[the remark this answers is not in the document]
esreli: Nice!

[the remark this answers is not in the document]
esreli: I'm thinking this through a bit more and it's not as straightforward as you suggest. Specifically, i'm running into an issue deriving a unique key from `Relationship.name`.  First, `Relationship.name` is optional which would require a fallback string that is both unique and end-user readable (because it is used by `localizedDescription`). Second, I believe it would be possible for two different tables to have the same table name. As such, we can't rely on `Relationship.name` as a unique hash. I do think it's worth incorporating some of your suggestion here but I don't think there's any way around performing a `map`.

data-collection/data-collection/Rich Popup/Attachments/RichPopupAttachmentsManager.swift
```diff
@@ -86,15 +86,23 @@ class RichPopupAttachmentsManager: AGSLoadableBase {
...
+        }
         
         stagedAttachments.append(stagedAttachment)
         
         return stagedAttachments.endIndex
     }
     
     
-    func deleteAttachment(at index: Int) -> Bool {
+    func deleteAttachment(at index: Int) throws -> Bool {
```
philium, line 101: Since this method now throws, could the return value be eliminated somehow? The way it is now, there are sort of two failure cases: when the method returns `false` or when it throws an error. Objective-C used to have the problem, which is why Apple translated Objective-C methods that take an `NSError **` error and return `Bool` into throwing methods that return `Void`.

esreli: TBH I kept this return value so I could use an `assertion` rather than a `precondition`- I want to "fail gracefully" but you're right, this and the other are programmer errors.

data-collection/data-collection/Rich Popup/Related Records/Relationships.swift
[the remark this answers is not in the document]
philium: Ah, I see. Alright! Thanks for checking.

philium (approved):

data-collection/data-collection/Rich Popup/Related Records/ManyToOneRelationship.swift
```diff
@@ -72,3 +72,12 @@ class ManyToOneRelationship: Relationship {
         }
     }
 }
+
+// MARK: Errors
+
+extension ManyToOneRelationship {
+    struct RequiredRelationship: LocalizedError {
+        let name: String
+        var errorDescription: String? { "Missing required relationship named \"\(name)\"." }
```
philium, line 81: A non-optional property can witness for an optional one, which I think would be more efficient. Maybe not? I don't know.

```suggestion
        var errorDescription: String { "Missing required relationship named \"\(name)\"." }
```

If you wanted to, you could use avoid needing to escape the quotation marks using extended string delimiters:

```suggestion
        var errorDescription: String { #"Missing required relationship named "\#(name)"."# }
```

data-collection/data-collection/Rich Popup/RichPopupManager.swift
```diff
@@ -749,3 +759,17 @@ extension RichPopupManager {
         }
     }
 }
+
+extension RichPopupManager {
+    
+    struct InvalidOperation: LocalizedError {
+        var errorDescription: String? { "The operation you are trying to perform is not permitted." }
```
philium, line 766:
```suggestion
        var errorDescription: String { "The operation you are trying to perform is not permitted." }
```

data-collection/data-collection/Rich Popup/RichPopupManager.swift
```diff
@@ -749,3 +759,17 @@ extension RichPopupManager {
...
+    }
+    
+    struct InvalidFields: LocalizedError {
+        let fields: [(name: String, error: Error)]
+        var errorDescription: String? {
```
philium, line 771:
```suggestion
        var errorDescription: String {
```

esreli:
@philium At first glance I thought the same you did in [this suggestion](https://github.com/Esri/data-collection-ios/pull/263#discussion_r530031783). Believe it or not, adherence to `LocalizedError` requires [`errorDescription`](https://developer.apple.com/documentation/foundation/localizederror/2946895-errordescription) to be declared as optional. Without doing so, a generic localized description is derived, ignoring the computed localized description i've submit in this PR. Here's how specifying the computed value as `String` vs `String?` fares when shown in an alert:

| var errorDescription: String? | var errorDescription: String |
| --- | --- |
| Simulator Screen Shot - iPhone 11 Pro - 2020-11-24 at 16 54 08 | Simulator Screen Shot - iPhone 11 Pro - 2020-11-24 at 16 55 03 |

I'll leave as is and will keep this requirement bookmarked- because i'll need to update other errors that conform to `LocalizedError`.

philium:
That is fascinating! That must be due to the default implementations provided. Normally it wouldn't be a problem to supply a non-optional property to satisfy the requirement for an optional one.

esreli:
https://github.com/Esri/data-collection-ios/pull/263#issuecomment-733399131
It is! I had operated under the same assumption and it took me some time to debug because it didn't cross my mind that the optional was strictly required. Though it is nice a default implementation is provided.

data-collection/data-collection/Rich Popup/RichPopupManager.swift
```diff
@@ -338,50 +351,50 @@ class RichPopupManager: AGSPopupManager {
...
-                    try parentManager.delete(oneToMany: richPopup)
-                }
-                catch {
-                    relatedRecordsErrors.append(error)
+        let errors = parentOneToManyManagers.keyEnumerator().allObjects
+            .compactMap { relationship -> (OneToManyRelationship, Error)? in
+                if let relationship = relationship as? OneToManyRelationship {
+                    if let parentManager = parentOneToManyManagers.object(forKey: relationship) {
+                        do {
+                            try parentManager.delete(oneToMany: richPopup)
```
philium, line 382: FYI: The Swift standard library has an [unwritten rule](https://forums.swift.org/t/why-does-the-implementation-of-uniqued-not-use-filter/41049/3) that they don't pass functions with side effects to methods like `filter`, `map`, etc. I think that's a good rule as the side effects can be easily missed when reading the code. I almost didn't see this when I first read this closure. You might consider whether adopting that rule could increase readability of this function.

data-collection/data-collection/Rich Popup/RichPopupManager.swift
```diff
@@ -395,29 +408,29 @@ class RichPopupManager: AGSPopupManager {
...
         
-        let foundManager = relationships.manyToOne.first { (manager) -> Bool in
-            
-            guard let relationshipInfo = manager.relationshipInfo else {
-                return false
+        let foundManager = relationships.manyToOne
+            .first { (manager) -> Bool in
+                guard let relationshipInfo = manager.relationshipInfo else { return false }
+                return relationshipInfo == info
             }
```
philium, line 430: This is more complicated than it needs to be. An optional value can be compared to a non-optional one using [`==`](https://developer.apple.com/documentation/swift/optional/2950146):

> Use this equal-to operator (==) to compare any two optional instances of a type that conforms to the Equatable protocol. The comparison returns true if both arguments are nil or if the two arguments wrap values that are equal. Conversely, the comparison returns false if only one of the arguments is nil or if the two arguments wrap values that are not equal.

```suggestion
            .first { $0.relationshipInfo == info }
```

data-collection/data-collection/Rich Popup/RichPopupManager.swift
```diff
@@ -525,22 +528,21 @@ class RichPopupManager: AGSPopupManager {
...
-            let relationship = foundManager,
-            let relatedFeature = relationship.popup?.geoElement as? AGSArcGISFeature,
-            let info = relationship.relationshipInfo
-            else {
-                throw RichPopupManagerError.cannotRelateFeatures
+        let foundManager = richPopup.relationships?.oneToMany
+            .first { (manager) -> Bool in
+                guard let relationshipInfo = manager.relationshipInfo else { return false }
+                return relationshipInfo == info
+            }
```
philium, line 539:
```suggestion
            .first { $0.relationshipInfo == info }
```

data-collection/data-collection/Rich Popup/RichPopupManager.swift
```diff
@@ -560,20 +562,20 @@ class RichPopupManager: AGSPopupManager {
...
-            return relationshipInfo == info
+        guard let feature = popup.geoElement as? AGSArcGISFeature else {
+            preconditionFailure("Unsupported record type.")
         }
+        
+        let foundRelationship = richPopup.relationships?.oneToMany
+            .first { (manager) -> Bool in
+                guard let relationshipInfo = manager.relationshipInfo else { return false }
+                return relationshipInfo == info
+            }
```
philium, line 573:
```suggestion
            .first { $0.relationshipInfo == info }
```

esreli:
@mhdostal @philium The old `AppError` design pattern is now removed from the app. This PR is ready for review. @philium thank you for all of your help thus far!

data-collection/data-collection/Extensions/ArcGIS/AGSArcGISFeature/AGSArcGISFeature+Collection.swift
```diff
@@ -26,21 +26,40 @@ extension Collection where Iterator.Element == AGSArcGISFeature {
     
     func asPopups() throws -> [AGSPopup] {
         
-        guard let first = first else {
+        guard count > 0 else {
```
philium, line 29: At first I was going to suggest using `isEmpty`, but then I noticed that what follows is just returning the result of a `map` operation. If `self` is empty, the result of the `map` will be an empty array. Therefore I would suggest removing this guard statement altogether.

data-collection/data-collection/Extensions/ArcGIS/AGSMapView/AGSMapView+ExtentForCGRect.swift
```diff
@@ -0,0 +1,54 @@
+//// Copyright 2017 Esri
```
philium, line 1: This file isn't in the project. Was it committed by mistake?

esreli: This extension was moved to the caller file, `MapViewController+LocationSelection.swift`. I'm not sure why it wasn't deleted from source, it was probably an Xcode user error I performed. Thank you.

mhdostal (approved):

PR #264, opened 2020-11-24
AppErrors: migrating Relationship conformance and general improvements

esreli:
This PR improves the type `Relationship` in a few ways.

- Establish `Relationship.MissingRelationship` error type
- Improves how Errors are managed by the class
- Improves `Relationship.queryAndSortAllRelatedPopups` to leverage swift Result type

### DIFF
--- a/data-collection/data-collection/Rich Popup/Related Records/Relationship.swift
+++ b/data-collection/data-collection/Rich Popup/Related Records/Relationship.swift
@@ -31,6 +31,11 @@ class Relationship: AGSLoadableBase {
             return nil
         }
         
+        guard let feature = popup.geoElement as? AGSArcGISFeature,
+              feature.featureTable is AGSArcGISFeatureTable else {
+            return nil
+        }
+        
         self.relationshipInfo = info
         self.relatedTable = table
         self.popup = popup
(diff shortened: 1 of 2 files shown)


data-collection/data-collection/Rich Popup/Related Records/Relationship.swift
```diff
@@ -58,39 +63,28 @@ class Relationship: AGSLoadableBase {
...
-            
-            guard let popups = popupsResults else {
-                self.loadDidFinishWithError(NSError.unknown)
-                return
+            else if let popups = popupsResults {
+                self.processRecords(popups)
+                self.loadDidFinishWithError(nil)
+            }
+            else {
+                preconditionFailure("Something went very wrong.")
```
philium, line 86: You might consider using `assert` here instead of `preconditionFailure` as this is more of a sanity check ("Use this function for internal sanity checks") than anything else. Although to be honest, at least in the samples app, we assume this case never happens as it _shouldn't_ ever happen.

philium (approved):

esreli:
Thank you @philium. I've applied your suggestion and am assuming the rest is okay. Merging.

PR #273, opened 2020-12-08
Post Pop-up Editing Bug

esreli:
This PR resolves the bug, ['Data Collection iOS, deleted pop-up remains in small pop-up view after delete is performed and full pop-up view is dismissed'](https://github.com/ArcGIS/open-source-apps/issues/482).

The solution uses a Combine `PassthroughSubject` to broadcast when edits are made. Subscribers are notified after an add/update/delete operation.


mhdostal (approved):
That is really slick.  Very nice!

PR #276, opened 2021-02-22
Mhd/504 image picker

mhdostal:
This adds an additional check for iOS 14 authorization status after the user selects the desired permissions in-app.  The result of picking "Selected Photos" as the new permissions is the same as if the user picked "None":

Screen Shot 2021-02-22 at 11 07 26 AM

### DIFF
--- a/data-collection/data-collection/Utilities/ImagePickerPermissions.swift
+++ b/data-collection/data-collection/Utilities/ImagePickerPermissions.swift
@@ -296,13 +296,16 @@ extension ImagePickerPermissions {
             PHPhotoLibrary.requestAuthorization { (status) in
                 DispatchQueue.main.async {
                     var granted: Bool
+                    var shouldOpenSettings: Bool
                     if #available(iOS 14, *) {
                         granted = PHPhotoLibrary.authorizationStatus(for: .readWrite) == .authorized
+                        shouldOpenSettings = !granted
                     }
                     else {
                         granted = PHPhotoLibrary.authorizationStatus() == .authorized
+                        shouldOpenSettings = false
                     }
-                    completion(granted, false)
+                    completion(granted, shouldOpenSettings)
                 }
             }
         }
--- a/data-collection/data-collection/Utilities/ImagePickerPermissions.swift
+++ b/data-collection/data-collection/Utilities/ImagePickerPermissions.swift
@@ -296,16 +296,13 @@ extension ImagePickerPermissions {
             PHPhotoLibrary.requestAuthorization { (status) in
                 DispatchQueue.main.async {
                     var granted: Bool
-                    var shouldOpenSettings: Bool
                     if #available(iOS 14, *) {
                         granted = PHPhotoLibrary.authorizationStatus(for: .readWrite) == .authorized
-                        shouldOpenSettings = !granted
                     }
                     else {
                         granted = PHPhotoLibrary.authorizationStatus() == .authorized
-                        shouldOpenSettings = false
                     }
-                    completion(granted, shouldOpenSettings)
+                    completion(granted, false)
                 }
             }
         }
--- a/data-collection/data-collection/Utilities/ImagePickerPermissions.swift
+++ b/data-collection/data-collection/Utilities/ImagePickerPermissions.swift
@@ -295,7 +295,14 @@ extension ImagePickerPermissions {
         func request(_ completion: @escaping (_ granted: Bool, _ shouldOpenSettings: Bool) -> Void) {
             PHPhotoLibrary.requestAuthorization { (status) in
                 DispatchQueue.main.async {
-                    completion(PHPhotoLibrary.authorizationStatus() == .authorized, false)
+                    var granted: Bool
+                    if #available(iOS 14, *) {
+                        granted = PHPhotoLibrary.authorizationStatus(for: .readWrite) == .authorized
+                    }
+                    else {
+                        granted = PHPhotoLibrary.authorizationStatus() == .authorized
+                    }
+                    completion(granted, false)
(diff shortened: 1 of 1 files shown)


data-collection/data-collection/Utilities/ImagePickerPermissions.swift
```diff
@@ -295,7 +295,14 @@ extension ImagePickerPermissions {
         func request(_ completion: @escaping (_ granted: Bool, _ shouldOpenSettings: Bool) -> Void) {
             PHPhotoLibrary.requestAuthorization { (status) in
                 DispatchQueue.main.async {
-                    completion(PHPhotoLibrary.authorizationStatus() == .authorized, false)
+                    var granted: Bool
```
esreli, line 298: This value will be established once and as such we can specify this variable `let`.
```suggestion
                    let granted: Bool
```

esreli (changes requested):
I think this PR is a good one. I have a few open questions remaining.

### API Parity?

Should the method on line `291` get the same treatment (the iOS 14 compatibility check and subsequent PHPhotoLibrary authorization status check) as is applied here?

### Selected Photos Permission Preference

AFAIK this PR doesn't fully solve for the case where a user has chosen 'Selected Photos' permission (see: [504](https://github.com/ArcGIS/open-source-apps/issues/504)). What do you think we should do in the case where the user has chosen 'Selected Photos'? I think the new image picker is off the table for now, do we have other options?

I was able to access the Selected Photos preference following these steps:
1. Launch application for the first time on a cleaned device/simulator running iOS 14
1. Identify tree feature
1. Expand tree feature to full pop-up view
1. Navigate to 'Attachments' pane
1. Tap 'Add Attachment'
1. Agree to start editing
1. Choose 'Select Photos...' permission

At this point an image picker is presented allowing me to choose 'selected photos'. Finishing the selection produces an error:

simulator_screenshot_8D9E66B1-C3A3-4173-86CD-4FAB8B34B5C2

The next time I try to add an attachment, i'm not told why I can't choose a photo despite having specified 'Selected Photos' permission.

simulator_screenshot_91C0958C-C2A9-4412-A56A-D0EC9F914820

#### Potential Solution

To me, there seems to be two problems here.

##### New Picker Automatic Presentation

The new picker is presented by the system automatically when we are trying to create a paradigm where it is unsupported. I've found a plist permission key that would prevent the picker from automatically presenting after the 'Selected Photos' permission is set.

Do I understand this problem correctly? I could be wrong here...

If this is the problem, we might be able to control when the picker is presented in code rather then the default system presentation behavior via the plist permissions key, [`PHPhotoLibraryPreventAutomaticLimitedAccessAlert`](https://developer.apple.com/forums/thread/650911).

##### Helpful Alert Dialog

I think we should consider crafting a dialog that helps the user understand why their 'Selected Photos' permission won't allow them to pick an image. We should probably tell them that permission isn't currently supported, changing the permission to 'All Photos' will allow them to add images from library, and that full support will come in the next release.

What do you think?

mhdostal:
@esreli Yes, I agree more work can be done to improve this.  Since it's going to be more involved, are you OK if we push this off until v1.3?

esreli:
> @esreli Yes, I agree more work can be done to improve this. Since it's going to be more involved, are you OK if we push this off until v1.3?

I think the extensive work can be put off yes. I do think we can use this opportunity to check the auth status for if the user has chosen 'selected photos' so that we can let them know why they don't have access to the picker when they thought they should. Am I making sense here?

mhdostal:
> Am I making sense here?

Yes, you are, I think your proposal is good.  I'll craft the alert and ping you when it's done.

mhdostal:
Here's the new alert.  I've made the code changes as well to check for iOS 14 and `.limited` status when displaying the `settings` alert.

Screen Shot 2021-02-23 at 12 23 53 PM

esreli (changes requested):
The alert is much better, thank you! Unfortunately, I am still experiencing the behavior described in the section named **Selected Photos Permission Preference** of my [last review](https://github.com/Esri/data-collection-ios/pull/276#pullrequestreview-595594879).

Perhaps this line from my last review might be helpful?

> If this is the problem, we might be able to control when the picker is presented in code rather then the default system presentation behavior via the plist permissions key, [`PHPhotoLibraryPreventAutomaticLimitedAccessAlert`](https://developer.apple.com/forums/thread/650911).

mhdostal:
> If this is the problem, we might be able to control when the picker is presented in code rather then the default system presentation behavior via the plist permissions key, [`PHPhotoLibraryPreventAutomaticLimitedAccessAlert`](https://developer.apple.com/forums/thread/650911).

I tried that and it had no affect for our use case; it still showed the `Select Photos...` options and displayed the image selection view when selected.  According to the docs: _"By default, the system automatically prompts the user to update their limited library selection once per app life cycle."_  This seems to indicate that it will prevent asking the user to update their selection, rather than preventing them from making the initial selection.

esreli:
> > If this is the problem, we might be able to control when the picker is presented in code rather then the default system presentation behavior via the plist permissions key, [`PHPhotoLibraryPreventAutomaticLimitedAccessAlert`](https://developer.apple.com/forums/thread/650911).
> 
> I tried that and it had no affect for our use case; it still showed the `Select Photos...` options and displayed the image selection view when selected. According to the docs: _"By default, the system automatically prompts the user to update their limited library selection once per app life cycle."_ This seems to indicate that it will prevent asking the user to update their selection, rather than preventing them from making the initial selection.

I guess there's no way around it then. Good thing we'll tackle this soon in `v.next-1.3`!

esreli (approved):

PR #278, opened 2021-03-04
V.next 1.2.3

mhdostal:
Updates the master branch with the v1.2.3 changes.  All of these changes have been reviewed; this is just a sanity check.


esreli (approved):
I think we're good to go here 👍

PR #280, opened 2021-03-04
Rebase 1.3 on 1.2.3

esreli:


mhdostal (approved):
This all looks good.  I compared this rebased code with merged code and accounted for all the differences, so the rebased code (this PR) is correct.

esreli:
And we're off!

PR #282, opened 2021-03-14
Support iOS 14 PHPickerViewController

esreli:
This PR introduces support for the iOS 14 `PHPickerViewController`. Summary of work:

- Removes `ImagePickerPermissions`
- Introduces os-availability check for `UIImagePickerController` vs `PHPickerViewController`
- Easier to read API

_Note, a user does not have to grant access to the `PHPhotoLibrary` in order to use `PHPickerViewController`. As such, no device authorization request is issued._


mhdostal (approved):
I'm surprised user permission isn't needed for the photo picker, but Ok.

Looks good to me.

esreli:
> I'm surprised user permission isn't needed for the photo picker, but Ok.
> 
> Looks good to me.

It took me a bit to wrap my head around as well. I think the idea is that because the content is derived from an item provider, we are programmatically closed off to the directory file system that would allow us to scour all of the user's photos.

PR #284, opened 2021-03-15
Progress view controller

esreli:
This PR removes the 3rd party dependency `SVProgressHUD` and replaces it with the home baked `ProgressViewController`. The main reason being `SVProgressHUD` is a third party dependency that we can no longer depend on. This became even more clear when iOS 13 deprecation warnings started screaming.

Included in this PR,

- Removes `SVProgressHUD` source code
- Removes objective-c bridging header file
    - Corrects implied `import` statements previously assumed by bridging file
- Introduces `ProgressViewController`


mhdostal (commented):
Are you doing anything special with the color of the ActivityIndicator?  When working with an online map, it appears to be green, like the title bar color.  When offline, it is still that shade of green, even though the map is in "offline mode" where the title bar is gray.

esreli:
> Are you doing anything special with the color of the ActivityIndicator? When working with an online map, it appears to be green, like the title bar color. When offline, it is still that shade of green, even though the map is in "offline mode" where the title bar is gray.

As of right now the color of the activity indicator is left untouched, using the UIAppearance protocol as a proxy for defining its tint color.

It's hard to say, I would suggest in a perfect world we actually would modify the app to allow light/dark mode to drive the color of the navigation bar and online/offline work mode to drive UI element tint color. What do you think?

mhdostal:
@esreli Thanks for the explanation.  I think it's OK for now. The activity indicator taking it's color from UIAppearance is fine; I don't think we need to do anything additional for light/dark mode at this time.

mhdostal (approved):

PR #287, opened 2021-03-22
Mhd/multiple identify results iii

mhdostal:
This adds support for Multiple Identify Results, shown in a Floating Panel.  Aside from showing all identify results and feature details/inspections/attachments in a floating panel, the functionality is identical to the previous (SmallPopupView) functionality


mhdostal:
@esreli  - I'd like your feedback on the current implementation:
- Does the UI/workflow match your expectations?
- Code organization feedback; I kept the main framework essentially the same as before, but just inserted the `IdentifyResultsViewController` at the beginning of the process.

There are still some outstanding items.  Here's what works and doesn't (yet):

What works:
- Viewing multiple identify results, both on and offline
- Viewing and editing details (and attachments) for each individual identify result
- Viewing, adding, deleting, and editing of inspection results (and attachments)

Remaining issues:
- Deleting feature does not remove it from list of identify results
- Creating new feature: after selecting geometry, crash while “Preparing new trees”.  Need to figure out out to jump directly to popups detail VC.
- Viewing feature details causes navigation bar to appear below floating panel header, but it seems out of place, as it is the ‘online green’ color. (doesn’t crash or anything, just looks somewhat out-of-place).  It *does* appear in the correctly color when going offline.  Your thoughts on this?
- When displaying identify results results in floating panel, the panel initial displays very large, then shrinks immediately to the ‘partial’ state, resulting in a brief flash on screen.  It needs to display immediately at correct size.
- In the Popup details view, the TableView section header for the first section (attributes) is being displayed, when it shouldn’t be.  It shows up as an empty space above the first feature attribute.
- When selecting a new identify result, need to update the selected feature on the map.

esreli:
> @esreli - I'd like your feedback on the current implementation:

I'll have a look shortly!

esreli (commented):
I think this is a great start. I can dive more into code and code architecture upon second pass, if you'd like- is that okay with you?

I have made a few in-line comments, one in particular references the use of subtitles.

Thanks!

### UI/UX

With the new floating panel design, the navigation bar now starts to feel a bit out of place. I think we draw off the design established in the .Net app for the iOS implementation as well except I think the mobile design should omit the `"< i of n >"` toggle UI control; the best part here is that we could move the bar button items from the navigation to the tool bar with likely very little friction. 

I drew up some prototypes to see how the iOS implementation might look following the same design as we have in .Net.

#### Pop-up View Mode

| .Net | iOS Prototype |
| --- | --- | 
| Screen Shot 2021-03-23 at 9 41 43 AM | Screen Shot 2021-03-23 at 11 44 48 AM |

#### Pop-up Edit Mode

| .Net | iOS Prototype |
| --- | --- | 
| Screen Shot 2021-03-23 at 9 42 00 AM | Screen Shot 2021-03-23 at 11 44 54 AM |

### Code Organization

I was only able to take a cursory glance at the code. It seems like a good start here. If you would like a more thorough code review, I'd be happy to give one.

You're also welcome to change and view architecture concepts as you see fit, the app is changing and the architecture can change alongside it if needed.

### Unknown UI Element

There appears to be a small white sliver across the top of the map view, just beneath the navigation bar. Digging into the view hierarchy, it appears the small pop-up view is nested up top behind the navigation bar. Is this intentional?

| Flat | Layer Hierarchy |
| --- | --- |
|  Screen Shot 2021-03-23 at 2 08 30 PM | Screen Shot 2021-03-23 at 2 07 48 PM |

mhdostal:
@esreli Thank you for the feedback, it all looks good.  I like the idea of the Toolbar as well.  Let me make the changes, including to the UI, and then submit it back to you for a more detailed review of the UI and code.

esreli:
> @esreli Thank you for the feedback, it all looks good. I like the idea of the Toolbar as well. Let me make the changes, including to the UI, and then submit it back to you for a more detailed review of the UI and code.

Sounds good 👍

esreli:
@mhdostal Also, @nCastle1 designed a tool that will generate Esri calcite icons for iOS. All of the icons included in the above screenshots were derived from the calcite icon set. You can follow [these instructions](https://github.com/Esri/calcite-ui-icons#build-png-icons-for-mobile-and-desktop) for rendering icons for iOS consumption.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+MapViewMode.swift
```diff
@@ -17,12 +17,37 @@ import UIKit
...
+                    // Set the selected popups on the identify results view controller.
+                    identifyResultsVC.selectedPopups = self.selectedPopups
+                    identifyResultsVC.popupChangedHandler = { [weak self] (richPopup: RichPopup?) in
+                        // Use the geometry engine to determine the nearest pop-up to the touch point.
+                        if let popup = richPopup {
+                            self?.setCurrentPopup(popup: popup)
+                        }
+                        else {
+                            self?.clearCurrentPopup()
+                        }
```
esreli, line 43: I'm unclear what the purpose is of this? Why does MapViewController need to maintain a reference to the current pop-up? Now that `RichPopupViewController` is floated, can't we offload the burden to that object?

data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
```diff
@@ -190,6 +184,13 @@ class MapViewController: UIViewController {
     // MARK: Current Pop-Up
     
     private(set) var currentPopupManager: RichPopupManager?
+    private(set) var selectedPopups = [RichPopup]()
+
+    func setSelectedPopups(popups: [RichPopup]) {
+        // Clear existing selection
+        (currentPopupManager?.popup.feature?.featureTable?.layer as? AGSFeatureLayer)?.clearSelection()
+        selectedPopups = popups
+    }
```
esreli, line 194: Do we need to maintain these references anymore?

data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
```diff
@@ -23,14 +23,12 @@ class MapViewController: UIViewController {
     enum MapViewMode: Equatable {
         case defaultView
         case disabled
-        case selectedFeature(featureLoaded: Bool)
+        case selectedFeature(visible: Bool)
         case selectingFeature
         case offlineMask
     }
```
esreli, line 29: Does this architecture hold up anymore? It was designed a while ago when the app was much more simple. You are absolutely welcome to scrap it in favor of a more robust architecture if you feel inclined to do so.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupViewController.swift
```diff
@@ -456,3 +439,27 @@ fileprivate extension UIViewController {
...
+            fpItem = top.floatingPanelItem
+        }
+        else {
+            fpItem = FloatingPanelItem()
+        }
+        
+        currentFloatingPanelItem = fpItem
+        return fpItem
+    }
+}
```
esreli, line 465: Do we have to derive the `FloatingPanelItem` from the rich popup details view controller? Since we have access to the same object, could we put `RichPopupViewController` in charge of deriving it's own floating panel item?

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupDetailsViewController/RichPopupDetailsViewController.swift
```diff
@@ -82,3 +82,14 @@ class RichPopupDetailsViewController: UITableViewController {
...
+extension RichPopupDetailsViewController: FloatingPanelEmbeddable {
+    var floatingPanelItem: FloatingPanelItem {
+        let floatingPanelItem = FloatingPanelItem()
+
+        let richPopup = popupManager.richPopup
+        floatingPanelItem.title = richPopup.title
+        floatingPanelItem.subtitle = richPopup.description
+        return floatingPanelItem
+    }
+}
```
esreli, line 95: I'm wondering if we should put `RichPopupViewController` in charge of maintaining it's `FloatingPanelItem` instead of offloading that responsibility to `RichPopupDetailsViewController`

data-collection/data-collection/View Controllers/Identify Results View Controller/IdentifyResultsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+        }
+        return cell
+    }
+    
+    var richPopup: RichPopup?
+
+    // MARK: - Navigation
+
+    // In a storyboard-based application, you will often want to do a little preparation before navigation
+    func prepareRichPopupViewController(_ richPopup: RichPopup) -> UIViewController? {
```
esreli, line 94: Is this method is used? If not, can we delete it?

esreli (changes requested):
Thanks Mark, I think you're off to a good start. What you have here is definitely a proof of concept and really fun to use. I'm glad it's starting to come together.

I'm offering feedback in three ways, **General UI** Feedback, General **Code** Feedback, and **In-line code** feedback.

## General UI Feedback

### Push/Pop Animation

The floating panel view controller push/pop animations slides the newly added or removed view in a top-to-bottom direction. Can we switch this to left-to-right?

## General Code Feedback

### `MapViewMode` Architecture

Does it hold up anymore? The architecture was designed when the requirements were much smaller and I think it's starting to show its weaknesses. For example, I can encounter bugs like so,

1. Identify features -> display features in list
2. Select feature
3. Select edit
4. Tap (`+`) add new feature button > the editing session disappears

Perhaps it's not worth investing the extra time here to soup up this architecture but we should be sure to account for all sorts of edge cases that can be reached now that the rich popup view controller is embed in the floating panel and the rest of the map view controller UI is still exposed.

### MapView Identified Results

While we aren't supporting the editing of geometries for any type, I do believe we should still display all found results in the `IdentifiedResultsViewController`. This is the functionality of the .Net build.

Would you update the map view identify code to capture all found features rather then only features with a point geometry type?

### Subtitles

Please see the previous PR comment on [subtitles](https://github.com/Esri/data-collection-ios/pull/287#discussion_r599934250).

data-collection/data-collection/View Controllers/Identify Results View Controller/IdentifyResultsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+                    // set the swatch into our dictionary and reload the row
+                    self?.swatches.setObject(image, forKey: symbol)
+                    tableView.reloadRows(at: [indexPath], with: .automatic)
+                })
+            }
+        }
+        return cell
+    }
+    
+    var richPopup: RichPopup?
```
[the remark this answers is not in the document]
mhdostal: Sounds good to me.

data-collection/data-collection/View Controllers/Identify Results View Controller/IdentifyResultsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+//                    print("[Error: RichPopup] relationships load error: \(error)")
+//                }
+//
+//                guard let self = self else { return }
+//
+//                self.populateContentIntoSmallPopupView(popup)
+//            }
+//        }
+//        else {
+//
```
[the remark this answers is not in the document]
mhdostal: Deleted.

data-collection/data-collection/View Controllers/Identify Results View Controller/IdentifyResultsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ArcGIS
+
+class IdentifyResultsViewController: UITableViewController, FloatingPanelEmbeddable {
+    var popupChangedHandler: ((RichPopup?) -> Void)?
```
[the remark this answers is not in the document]
mhdostal: I updated the `popupChangedHandler` callback in the MapViewController to also control the feature selection on the Map.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController+MapViewMode.swift
mhdostal: When creating a new feature, the FloatingPanel is not used (the view is presented modally).  The current popup methods have been updated a bit to handle feature selection on the map view as well.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
mhdostal: Yes, `selectedPopups` is set by the identify operation and then passed to the identify results vc.

data-collection/data-collection/View Controllers/Identify Results View Controller/IdentifyResultsViewController.swift
```diff
@@ -0,0 +1,170 @@
...
+            //                setCurrentPopup(popup: newPopup)
+            destination.popupManager = RichPopupManager(richPopup: newPopup)
+            destination.setEditing(true, animated: false)
+            //                mapViewMode = .selectedFeature(visible: false)
+        }
+        else if let popupManager = EphemeralCache.shared.object(forKey: .newRelatedRecord) as? RichPopupManager {
+            destination.setEditing(true, animated: false)
+            destination.popupManager = popupManager
+            destination.shouldLoadRichPopupRelatedRecords = false
+        }
```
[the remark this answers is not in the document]
mhdostal: Nope.  Deleted.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupViewController.swift
mhdostal: Do you mean just having the `RichPopupViewController` derive from `FloatingPanelEmbeddable` and not having `RichPopupDetailsViewController` derive from it?

data-collection/data-collection/View Controllers/Identify Results View Controller/IdentifyResultsViewController.swift
mhdostal: No, it's not used.  Nice catch.  Removed.

data-collection/data-collection/View Controllers/Rich Popup View Controllers/Rich Popup View Controller/RichPopupDetailsViewController/RichPopupDetailsViewController.swift
mhdostal: Everything is now handled inside the `RichPopupViewController`.

data-collection/data-collection/View Controllers/Identify Results View Controller/IdentifyResultsViewController.swift
```diff
@@ -0,0 +1,127 @@
...
+                    guard cell.tag == indexPath.hashValue, let image = image else { return }
+
+                    // set the swatch into our dictionary and reload the row
+                    self?.swatches.setObject(image, forKey: symbol)
+                    tableView.reloadRows(at: [indexPath], with: .automatic)
+                })
+            }
+        }
+        if let subtitle = subtitles.object(forKey: richPopup) {
+            cell.detailTextLabel?.text = subtitle as String
```
esreli, line 89:
```suggestion
        if let subtitle = subtitles.object(forKey: richPopup) as String? {
            cell.detailTextLabel?.text = subtitle
```

data-collection/data-collection/View Controllers/Identify Results View Controller/IdentifyResultsViewController.swift
```diff
@@ -0,0 +1,127 @@
...
+        }
+        else {
+            richPopup.evaluateSubtitle { [weak self] (subtitle) in
+                self?.subtitles.setObject(subtitle as NSString, forKey: richPopup)
+                self?.tableView.reloadRows(at: [indexPath], with: .automatic)
```
esreli, line 94: It's probably not something we need to change but in case the indexPath has changed (i.e. the table is reloaded with changes) between when the subtitle evaluation started and when the async function finished, we have the possibility of setting a subtitle to the wrong row or even a missing row. Just something to think about.

data-collection/data-collection/View Controllers/Map View Controller/MapViewController.swift
```diff
@@ -250,7 +268,7 @@ class MapViewController: UIViewController {
                 case .failure(let error):
                     self?.showError(error)
                 case .success(_):
-                    self?.refreshCurrentPopup()
+                    print("Feature added successfully")
```
esreli, line 271: I don't think we want to print this to console, can we `break` instead?

esreli (changes requested):
Functionally, I think we're almost there! It's very fun to use this tool. I've consulted the PR's description for what to review.

I've divided this review into three sections, 1) Addressing your remaining items found in the PR description, 2) general feedback, 3) in-line code feedback.
 
## Your remaining items

### Display large then shrinks to partial height

In your PR description, you outline this remaining issue,

> When displaying identify results results in floating panel, the panel initial displays very large, then shrinks immediately to the ‘partial’ state, resulting in a brief flash on screen. It needs to display immediately at correct size.

There's so much layout and transition code in the floating panel source files that I'm having a difficult time following all that's going on. I can take a deeper look after DS2021, if it's still unresolved.

My intuition on what's happening is that the view is [added](https://github.com/Esri/data-collection-ios/blob/mhd/MultipleIdentifyResults_III/data-collection/data-collection/View%20Controllers/Floating%20Panel%20View%20Controller/FloatingPanelController.swift#L829-L830) to the subview at full size and then the constraints for partial are calculated and the view is redrawn? Does that help at all?

### Table attributes section header

In your PR description, you outline this remaining issue,

> In the Popup details view, the TableView section header for the first section (attributes) is being displayed, when it shouldn’t be. It shows up as an empty space above the first feature attribute.

I think it's totally fine to leave as-is. In fact, I like the extra whitespace!

### Disappearing subtitle label

In your PR description, you outline this remaining issue,

> When viewing the list of identify results, the subtitle of the header view is visible. When viewing the feature details, there is no subtitle, so the header contracts; when popping back to the list of results, the header does not expand to display the subtitle again.

This is happening when the subtitle label's text value is empty or null. Why don't we block the subtitle label from disappearing all together?

We have an [extension](https://github.com/Esri/data-collection-ios/blob/mhd/MultipleIdentifyResults_III/data-collection/data-collection/Extensions/UIKit/UILabel/UILabel%2BEmptyString.swift#L25) on `UILabel` that accounts for empty strings, inserting a space (" ") when a string is empty so that the label remains rendered within the stack view, but doesn't show any value.

We should update that extension to also capture nil strings,

```swift
func considerEmptyString() {
    if text == nil {
        text = " "
    }
    else if let labelText = text, labelText.isEmpty {
        text = " "
    }
}
```

And then you can use it in the Floating Panel Header Controller in both (2) times where we set the subtitle label's text value,

```swift
subtitleLabel.text = item.subtitle
subtitleLabel.considerEmptyString()
```

It's a bit hacky but it should resolve it.


## My general notes

### Creating new feature

I don't know how I didn't notice this until now but the UI for creating a new feature is not embed in the floating panel. The design of DC 1.3 is supposed the remove the use of the small pop-up view and remove the modal presentation of the rich pop0up view controller, opting instead to use the floating panel for all pop-up viewing and editing.

What remains here is the use of modal presentation of a rich popup view controller for the create new feature workflow.  Is it a big undertaking to embed the rich popup view controller into the floating panel for the create feature workflow?

mhdostal:
> What remains here is the use of modal presentation of a rich popup view controller for the create new feature workflow. Is it a big undertaking to embed the rich popup view controller into the floating panel for the create feature workflow?

The main driver of viewing popups in the floating panel is the IdentifyResults View.  It, and the navigation controller, really manage the pushing/popping of the individual identify results popups.  So bypassing that and displaying the RichPopupViewController as the topmost vc is presenting challenging in terms of dismissing the floating panel.  So it's going to take a little bit more time to figure out all the details.

@esreli In the meantime, can you review the current changes based on your feedback to make sure it's what you expected?  Thanks.

esreli:
@mhdostal will do! Thanks for looking in to this.

data-collection/data-collection/AppRules.swift
```diff
@@ -33,7 +32,6 @@ class AppRules {
         guard
             layer.isVisible,
             let featureTable = layer.featureTable,
-            featureTable.geometryType == .point,
             featureTable.isPopupActuallyEnabled else {
                 return false
         }
```
esreli, line 37: I'm still erring on the belief that this method is no longer needed. I think we should let the `mapView.identify` method drive what pop-ups are returned without the need to filter using this method. I'm also betting the results here are the same with or without this rule.

esreli (commented):
I think we're nearly there! Not much to report from my end. This tool is really fun to use!

data-collection/data-collection/AppRules.swift
mhdostal: Ok.  I've removed it and things still work as expected.

mhdostal:
@esreli I've made the final(?) changes.  I've put the 'add new feature' UI into a floating panel and resolved the remaining issues from your review.  Please re-review.

esreli (commented):
Functionally, I think we're nearly ready to go here. It's really coming together nicely!

You've identified a known bug, [**Potential loss of input data during feature creation/edit**](https://github.com/ArcGIS/open-source-apps/issues/568). Do you hope to tackle this in this PR or elsewhere?

### Unused images

There are few images that are added or changed in this PR that I can't find where they are referenced in doc. Can you determine if these images are used in doc or elsewhere?

- `docs/images/anatomy-floating-panel-view.png` 
- `docs/images/featured-img.png`

### Update doc TOC

Can you be sure to update the TOC for both README.md doc pages? You can do so using [this tool](https://github.com/Esri/data-collection-ios/tree/v.next#mdtoc).

mhdostal:
@esreli I've made all the changes.  The fix for [**Potential loss of input data during feature creation/edit**](https://github.com/ArcGIS/open-source-apps/issues/568) was a bit more involved and required:

- adding `start/endEditing` notifications
- updating the `isEnabled` property of the `Profile`, `Add New Feature`, and `Extras` button whenever editing started/ended.
- add a check for `isEditing` when determining whether to query on user map touch event.

I considered using another Combine Publisher (or expanding the current `EditsMade` one), but it was more foolproof to use  notifications.

Please re-review.

esreli (approved):
Great work here, it's feeling polished! I have just a few notes, feel free to take from them what you want and then I think we're ready to merge. Great work!

### Full Pop-up View Subtitle

The subtitle is generated for the list and works well!

Screen Shot 2021-05-04 at 10 55 23 AM

I think we could take the same subtitle and place it in the floating panel header view subtitle label. Thougts?

Screen Shot 2021-05-04 at 10 49 00 AM

### AppContext 

Just a small note on the use of the AppContext and notification center for broadcasting the rich popup view editing state.  Personally, I would have brokered a local relationship between the MapViewController and FloatingPanelViewController, allowing the two of them to coordinate how this state change together. Because the AppContext is only used for storing Notifications, it seems as if we could forego the AppContext all together. I would consider placing the Notifications stored in AppContext instead in the RichPopupViewController.

mhdostal:
@esreli I've made the following changes, per your review

- added subtitle to the details view controller
- instead of checking for `isEditing` in the map view touch delegate (to prevent identifying new features when editing), I just set the `mapView.touchDelegate = nil` when editing in `adjustUIForEditing`.  This is because `currentPopupManager.isEditing` is false when adding a new inspection record.  Swapping out the touchDelegate works in all cases.
- Moved the `didStartEditingNotification/didCancelEditingNotification ` notifications to `RichPopupViewController`.

> Personally, I would have brokered a local relationship between the MapViewController and FloatingPanelViewController, allowing the two of them to coordinate how this state change together.

Ideally, I agree this would have been the way to go.  However, the `MapViewController` only interfaces with the `IdentifyResultsViewController` directly (via the `FloatingPanel`).  The `IdentifyResultsViewController` is then responsible for hiding/showing the `RichPopupViewController`, which does NOT have a relationship with the `MapViewController` and there was no way to elegantly create one.  Hence the notification system.

esreli (approved):
This all works exactly as expected. Nice work!

PR #288, opened 2021-04-08
Mhd/multiple results doc

mhdostal:
Doc and screenshot changes to support Floating Panel and Multiple Identify Results.  Merging into "MulitpleIdentifyResults_III" as it contains all the MIR changes.


RELEASE.md
```diff
@@ -13,6 +13,10 @@
 - Fixes bug where `AddressLocator.onlineLocator` issues an authentication challenge amid creating a new feature.
 - Introduces support for iOS 14 PHPickerViewController.
 - Removes `SVProgressHUD` dependency, introduces `ProgressViewController` global presenter.
+- Introduces `FloatingPanelController` for displaying information in a customizable panel which "floats" about the map.
+- Moves the Bookmarks and Layer views into a `FloatingPanelController`.
```
esreli, line 17:
```suggestion
- Moves the **Bookmarks** and **Layers** views into a `FloatingPanelController`.
```

RELEASE.md
```diff
@@ -13,6 +13,10 @@
...
+- Adds support for displaying multiple identify results from all identifiable layers in a map.
+- Fixes Dark Mode support in the `RichPopupViewController`.
```
esreli, line 19:
```suggestion
- Adds Dark Mode support in the `RichPopupViewController`.
```

docs/README.md
```diff
@@ -369,23 +367,13 @@ The app operates on a set of rules driven by map definitions and pop-up configur
...
-
-### Small pop-up view rules
-
-After the identify function returns a single successful result, the app selects the result on the map and populates a small pop-up view (contained by a `ShrinkingView`, see the section entitled [_Custom Views_](#custom-views)).
-
-The small pop-up view prioritizes related record content over content derived from its own attributes.
+A tap gesture on the map view performs an identify function where only results for layers that adhere to certain rules are considered. These rules ask that the layer is visible and pop-ups are enabled. These rules are wrapped conveniently into a static class named `AppRules`.
 
-![Small Pop-up View](/docs/images/anatomy-small-popup-view.png)
+### Floating panel view rules
```
esreli, line 372: The shrinking view had a unique ruleset that defined what was shown. I would say it's more appropriate to modify the title this way. (If you do, don't forget to update the TOC!)
```suggestion
### Floating panel view results
```

docs/README.md
```diff
@@ -195,13 +195,11 @@ You can select the Layers item in Extras to view the symbology for each layer. Y
 
 ### Identify map features
 
-Tapping the map performs an identify function on the map. One best result is chosen, a small pop-up view is revealed and the feature is selected on the map. If no results are found, the user is notified.
+Tapping the map performs an identify function on the map. One results are identified, a floating panel view is displayed and the feature(s) are selected on the map. If no results are found, the user is notified.
```
esreli, line 198:
```suggestion
Tapping the map performs an identify function on the map. After results are identified, a floating panel view is displayed and the feature(s) are selected on the map. If no results are found, the user is notified.
```

docs/README.md
```diff
@@ -219,11 +217,11 @@ The app supports pop-ups with related records (related tables) and attachments.
 
 ### View and edit data with pop-ups
 
-After identifying a pop-up, tapping the small pop-up view modally presents that pop-up in a more detailed pop-up view.
+After identifying a pop-up, tapping a result presents that pop-up in a more detailed pop-up view.
 
 #### View a pop-up
 
-A full screen table-based view controller allows the user to interrogate the map view's selected pop-up in greater detail. The table-based view is broken down into a number of sub-components.
+A table-based view controller is displayed in a floating panel which allows the user to interrogate the map view's selected pop-up in greater detail. The table-based view is broken down into a number of sub-components.  The floating panel can be resized by dragging the handle at the top (iPhone) or bottom (iPad) of the view.
```
esreli, line 224: Remind me, is it iPhone vs iPad? or compact vs regular width? Does it matter?

docs/README.md
```diff
@@ -251,7 +249,7 @@ The pop-up's attributes configured as editable can be edited and validated inlin
 
 As values for fields are updated, the app informs the user of invalid changes and why it's invalid. The pop-up won't save if there are invalid fields.
 
-Edits can be discarded by tapping 'Cancel' in the navigation bar. Saving the changes requires every field to pass validation and can be committed by tapping 'Done' in the navigation bar.
+Edits can be discarded by tapping the 'X' button in the bottom tool bar. Saving the changes requires every field to pass validation and can be committed by tapping the 'Disk' button in the bottom tool bar.
```
esreli, line 252:
```suggestion
Edits can be discarded by tapping the 'X' button in the bottom tool bar. Saving the changes requires every field to pass validation and can be saved by tapping the 'Disk' button in the bottom tool bar.
```

docs/README.md
```diff
@@ -369,23 +367,13 @@ The app operates on a set of rules driven by map definitions and pop-up configur
...
 ### Map identify rules
 
-A tap gesture on the map view performs an identify function where only results for layers that adhere to certain rules are considered. These rules ask that the layer is visible, is of point type geometry and pop-ups are enabled. These rules are wrapped conveniently into a static class named `AppRules`.
-
-### Small pop-up view rules
-
-After the identify function returns a single successful result, the app selects the result on the map and populates a small pop-up view (contained by a `ShrinkingView`, see the section entitled [_Custom Views_](#custom-views)).
-
-The small pop-up view prioritizes related record content over content derived from its own attributes.
+A tap gesture on the map view performs an identify function where only results for layers that adhere to certain rules are considered. These rules ask that the layer is visible and pop-ups are enabled. These rules are wrapped conveniently into a static class named `AppRules`.
```
esreli, line 370: With the introduction of all geometry types, is this section really needed anymore? For that matter, is [AppRules. isLayerIdentifiable](https://github.com/Esri/data-collection-ios/blob/mhd/MultipleResults_Doc/data-collection/data-collection/AppRules.swift#L31-L40) still needed?

esreli (approved):
Just a few notes in-line and this is looking very good 👍

docs/README.md
mhdostal: Yes, `isLayerIdentifiable` is still needed as it still checks to see if the layer is visible and if it has popups enabled.

docs/README.md
mhdostal: You are correct, it should be `compact` vs. `regular` width.

docs/README.md
esreli: > to see if the layer is visible

Do you know if this behavior is also the default?

> if it has popups enabled

We wrote this "is actually enabled" subset to ensure we get results that we can depend on for the point-only workflow and our dataset. 

I wonder if we should consider simply allowing the SDK to return results based on the default identify operation, should we?

docs/README.md
```diff
@@ -221,7 +221,7 @@ After identifying a pop-up, tapping a result presents that pop-up in a more deta
 
 #### View a pop-up
 
-A table-based view controller is displayed in a floating panel which allows the user to interrogate the map view's selected pop-up in greater detail. The table-based view is broken down into a number of sub-components.  The floating panel can be resized by dragging the handle at the top (iPhone) or bottom (iPad) of the view.
+A table-based view controller is displayed in a floating panel which allows the user to interrogate the map view's selected pop-up in greater detail. The table-based view is broken down into a number of sub-components.  The floating panel can be resized by dragging the handle at the top (for a compact width size class) or bottom (for a regular width size class) of the view.
```
esreli, line 224:
```suggestion
A table-based view controller is displayed in a floating panel which allows the user to interrogate the map view's selected pop-up in greater detail. The table-based view is broken down into a number of sub-components.  The floating panel can be resized by dragging the handle at the top (for a compact-width size class) or bottom (for a regular-width size class) of the view.
```

esreli (approved):
I think we're good here! Thanks!

docs/README.md
mhdostal: I originally had it with hyphens, but after reading the [Apple Doc](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout), I took out the hyphens.

docs/README.md
mhdostal: We may not need it for our situation, now that we allow identifying all geometry types, but I think it is still a useful point of customization for clients.

PR #290, opened 2021-04-15
Mhd/fix UI table view alert for layout outside view hierarchy

mhdostal:
In the console I'm seeing a warning for `UITableViewAlertForLayoutOutsideViewHierarchy`.  This happens when a `layoutIfNeeded` call is applied to a view not yet added to a superview.


esreli (approved):
Nice! This has been bugging me forever.

PR #292, opened 2021-04-16
Floating Panel Animation

esreli:
This simple PR sets the floating panel's initial frame off the bottom of the screen creating an effect where the floating panel animates upward or from the top of the screen if non-compact.

Note, I know this is a cheap solution!


mhdostal (approved):
That looks awesome, thanks!

PR #294, opened 2021-05-04
Fix for not prompting user to sign in when opening private map.

mhdostal:
This adds back in a "sign in alert" for use when attempting to load a private web map.  This functionality existed v1.2.3, but was removed (inadvertently?) when moving to the `GlobalAlertQueue`.


esreli (approved):
I was trying to figure out why I made this change. The best git blaming I can do is that it was lumped in to the huge architectural overhaul I did a while back and that I simply didn't offer the ability to sign in. 

Thanks for finding this!

PR #295, opened 2021-05-04
Use adaptive colors when showing enabled/disabled layer labels.

mhdostal:
Fixes the issue with label colors in the Layers View in dark mode.  Now it uses adaptive colors and defines static values for those colors in the `LayerCell` object.

### DIFF
--- a/data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
+++ b/data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
@@ -78,6 +78,9 @@ class LayerCell: UITableViewCell {
     @IBAction func visibilityChanged(_ sender: Any) {
         delegate?.visibilityChanged(self)
     }
+    
+    static let labelColor: UIColor = .label
+    static let dimmedColor: UIColor = .secondaryLabel
 }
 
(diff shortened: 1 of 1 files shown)


data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
```diff
@@ -78,6 +78,9 @@ class LayerCell: UITableViewCell {
     @IBAction func visibilityChanged(_ sender: Any) {
         delegate?.visibilityChanged(self)
     }
+    
+    static let labelColor: UIColor = .label
+    static let dimmedColor: UIColor = .secondaryLabel
```
esreli, line 83: We could probably remove these static variables and simply `set` the static colors they represent (`.label` and `.secondaryLabel`) in the two areas where they are used below.

esreli (approved):
Thanks for catching this! Just a simple suggestion, feel free to take it or leave it. Looks good!

data-collection/data-collection/View Controllers/Layer Contents View Controllers/LayerContentsTableViewController.swift
mhdostal: I wanted them in one place so any changes wouldn't have to be made in two places.

PR #296, opened 2021-05-04
Swift Package Manager support

mhdostal:
This adds both the Toolkit and SDK as Swift Packages.  It removes the old references to the SDK framework and the Toolkit and removes the Toolkit submodule, as it's not needed anymore.


esreli (approved):
This looks good to me. Would you note the change in the RELEASE notes?

esreli (approved):

PR #299, opened 2021-05-06
Fix for [iOS] Can't "Cancel" adding an attachment from Image Library (iOS 14) #573

mhdostal:
The one delegate method for the `PHPickerViewController` is designed to signify both Cancel and Accept operations.  Make sure we handle these two cases.


mhdostal:
> Since this branch doesn't have the attachments source type on iPad bug fix, I wasn't able to confirm this on iPad but it does work as expected on iPhone.

@esreli After merging, I tested this on an iPad and it works correctly.

